PR #2, opened 2016-06-10
add 'reviewed by' and timestamp to interface

MusikAnimal:


MusikAnimal:
going to guess that failure was not introduced by this PR? https://travis-ci.org/Niharika29/PlagiabotWeb/builds

src/Controllers/CopyPatrol.php
```diff
@@ -156,7 +160,7 @@ public function getReportLink( $ithenticateId ) {
 	 */
 	public function formatTimestamp( $datetime ) {
 		$datetime = strtotime( $datetime );
-		return date( 'd-m-y h:m:s', $datetime );
+		return date( 'Y-m-d h:m', $datetime );
```
MusikAnimal: Assuming this was fine, going off of the [wireframe](https://phab.wmfusercontent.org/file/data/7i6yksmlqna7ri4mywe5/PHID-FILE-abfr7ggyussiagpan7a3/copy_patrol_3_-_review_selected.jpg)

public_html/templates/index.html
```diff
@@ -1,100 +1,118 @@
...
-                            <div class="compare-website compare-pane">
-                                Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget
-                                dolor.
-                            </div>
-                        </div>
-                    {% endfor %}
-                {% endif %}
             </div>
         {% endfor %}
     {% endif %}
```
MusikAnimal: In short what happened here: I added a wrapper around the two divs that make up an entry in the interface, moving the padding and border styles to it, and indented everything in between. Much smaller change than it appears to be!

src/Controllers/Review.php
```diff
@@ -36,7 +36,14 @@ public function __construct( \Slim\Slim $slim = null ) {
 	protected function handleGet() {
 		$id = $this->request->get( 'id' );
 		$val = $this->request->get( 'val' );
-		$ret = $this->dao->insertCopyvioAssessment( $id, $val );
+
+		// TODO: disallow anonymous reviewers (likely by hiding the review buttons altogether)
+		$userData = $this->authManager->getUserData();
+		$user = $userData ? $userData->getName() : NULL;
+
+		$timestamp = gmdate( 'c' );
```
kaldari: This could use a comment, like "Get current UTC time as ISO 8601 timestamp."

src/Controllers/CopyPatrol.php
Niharika29: Ugh, this is the US vs rest-of-the-world debate all over. :P

src/Controllers/CopyPatrol.php
```diff
@@ -108,6 +108,10 @@ protected function handleGet() {
 				$records[$key]['editor_page_dead'] = false;
 				$records[$key]['editor_talk_dead'] = false;
 			}
+			if ( $record['status_user'] ) {
+				$records[$key]['reviewed_by_url'] = $this->getUserPage( $record['status_user'] );
+			}
+			$records[$key]['review_timestamp'] = $this->formatTimestamp( $record['review_timestamp'] );
```
Niharika29: Shouldn't this be in the if statement as well?

public_html/templates/index.html
```diff
@@ -1,100 +1,118 @@
 {% extends "base.html" %}
+{% import _self as macros %}
+
+{% macro reviewer_node( row ) %}
+    <div class="status-div--reviewer">
```
Niharika29: Any special reason for using '--' instead of just '-'?

public_html/css/index.css
```diff
@@ -154,3 +154,8 @@ h1 {
     /*border: 1px solid green;*/
     display: inline-block;
 }
+
+.status-div--reviewer {
+    margin-top: 10px;
+    white-space: nowrap;
+}
```
Niharika29: I think we should have this in a smaller font. Same for timestamp. I have diff_timestamp and edit count wrapped in <small> but doing this in CSS is more elegant, definitely.

src/Controllers/Review.php
```diff
@@ -36,7 +36,14 @@ public function __construct( \Slim\Slim $slim = null ) {
 	protected function handleGet() {
 		$id = $this->request->get( 'id' );
 		$val = $this->request->get( 'val' );
-		$ret = $this->dao->insertCopyvioAssessment( $id, $val );
+
+		// TODO: disallow anonymous reviewers (likely by hiding the review buttons altogether)
```
Niharika29: You can get rid of this TODO. I'm taking care of this in the route handler itself.

Niharika29:
The patch looks good. I made some inline comments. One thing I wanted to do here was to instantly display this information on the page when the review is made.  This should be fairly straightforward because I am already using AJAX to echo back a 'true', 'false' or 'Unauthorized'. You can check if the return value is true and fade-in the reviewer information (index.js). The base template already has the 'user' value which you can use and use the current timestamp. You don't need to echo these values back from the database.  
Also I made some important changes to the master branch so you might want to rebase. :)

src/Controllers/CopyPatrol.php
MusikAnimal: Well, for now anyway I figure since we're allowing unauthenticated users to perform reviews, we should at least record the timestamp. Once we require authentication we could probably get rid of the if statement altogether

public_html/templates/index.html
MusikAnimal: [BEM](http://getbem.com/) methodology, which I tend to do instinctively. Going by the [MediaWiki coding conventions](https://www.mediawiki.org/wiki/Manual:Coding_conventions/CSS) it looks like we don't do this, so I can unlearn it :)

public_html/css/index.css
MusikAnimal: A smaller font would avoid the username from potentially going offscreen, since the `nowrap` keeps it from wrapping. I added that rule because for most usernames it's going to make it look nicer (`MusikAnimal (WMF)` shows on two lines rather than one), but I was thinking we might also have an ellipsis; e.g. `Reviewed by ReallyLongUsernameBlah...` 

I definitely prefer CSS over allowing the browser to style things, but here since we're using Bootstrap it's applying explicit sizing with CSS, so we might consider using `<small>` to be acceptable.

src/Controllers/CopyPatrol.php
MusikAnimal: [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) = international standard!

src/Controllers/CopyPatrol.php
Niharika29: Nope, we aren't letting unauthenticated users perform reviews anymore, per the latest code in master. You can get rid of it now.

public_html/js/index.js
```diff
@@ -1,35 +1,42 @@
...
+				if ( ret.error === 'Unauthorized' ) {
+					alert( 'You need to be logged in to be able to review.' );
+				} else {
+					alert( 'There was an error in connecting to database.' );
+				}
+				$( buttonId ).addClass( 'btn-' + buttonClass ).removeClass( 'btn-' + buttonClass + '-clicked' ).blur();
+				$( unusedButtonId ).removeClass( 'btn-secondary' ).prop( 'disabled', false ).addClass( 'btn-' + unusedButtonClass );
+			}
+		}
+	);
```
MusikAnimal: Made spaces into tabs per [coding conventions](https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Whitespace) (unclear if they apply here, but figured they should)

src/Controllers/CopyPatrol.php
```diff
@@ -108,6 +108,11 @@ protected function handleGet() {
 				$records[$key]['editor_page_dead'] = false;
 				$records[$key]['editor_talk_dead'] = false;
 			}
+			$records[$key]['reviewed'] = $record['status_user'] ? true : false;
+			if ( $records[$key]['reviewed'] ) {
+				$records[$key]['reviewed_by_url'] = $this->getUserPage( $record['status_user'] );
+				$records[$key]['review_timestamp'] = $this->formatTimestamp( $record['review_timestamp'] );
+			}
```
MusikAnimal: Still need the `if` as there are some historical records with no reviewer

src/Controllers/Review.php
```diff
@@ -23,7 +23,7 @@
 
 use Wikimedia\Slimapp\Controller;
 
-class Review extends Controller {
+class Review extends CopyPatrol {
```
MusikAnimal: needed access to `getUserPage` and `formatTimestamp`

MusikAnimal:
Alright, here's round two. This brings some improvements, but one thing I noticed is that if I'm logged out the buttons give no indication that I need to log in. I think we should either hide the buttons from view, or allow them to be clicked and show the alert message that is currently coded in `index.js`. I guess this would mean moving the check for authentication either to `PlagiabotDao#insertCopyvioAssessment` or `Review#handleGet`

public_html/templates/index.html
```diff
@@ -1,100 +1,116 @@
 {% extends "base.html" %}
+{% import _self as macros %}
+
+{% macro reviewer_node( row ) %}
+    <div class="status-div-reviewer{{ row.reviewed ? '': ' none' }} status-div-reviewer-{{ row.ithenticate_id }}">
+        Reviewed by
+        <span>
+            <a class="reviewer-link" href="{{ row.reviewed_by_url }}" target="_blank">{{ row.status_user }}</a>
+        </span>
```
Niharika29: Why span above and div below?

MusikAnimal: the `<div>` will appear on it's own line, while `<span>` will be inline

PR #4, opened 2016-06-16
Add Less CSS preprocessing engine, some refactoring

MusikAnimal:
**Why use Less?**
- Our Twig templates have conditionals to add the proper CSS classes on individual elements based on the triage state, the same is true with the JavaScript in `saveState`
- To accomplish what we had in mind for the [undo review task](https://phabricator.wikimedia.org/T134597), we'd end up having to copy/paste Bootstrap styles, or use several lines of JavaScript with unnecessary event listeners to accomplish what should be a simple thing
- With Less, we can add a single CSS class name to `.status-div` that represents the triage state. From there we can use Bootstrap.less (not included in this PR, but easy to add) to extend the necessary classes and get the styles we want. So we have the initial state, but on `:hover` we just extend a different Bootstrap class. A few lines of code instead of a bunch of lines of code that need refactoring, but we can't refactor because native CSS prevents any attempt at being DRY
- There is little overhead considering the `index.css` file is cached, and is only regenerated when the Less files change (so once per deploy)
- You can use normal ole CSS in Less. No need to hesitate and feel like you need to read the docs. Just write and we can refactor as needed.
- MediaWiki does it, and much in the same way. This isn't really a compelling reason to use it on Tool Labs, but it'd at least be familiar to those comfortable with the MediaWiki frontend
- CSS preprocessors are just awesome :) They are incredibly easier to work with (nesting alone will win you over), and they can bring a huge CSS file down to a fraction of the size, making it much more maintainable


src/App.php
```diff
@@ -198,6 +199,16 @@ function () use ( $slim ) {
...
 			} );
+		$slim->get( '/index.css',
+			function () use ( $slim ) {
+				// Compile LESS if need be, otherwise serve cached asset
+				// Cached files get automatically deleted if they are over a week old
+				$lessFiles = array( APP_ROOT . '/src/Less/index.less' => '/copypatrol/' );
+				$options = array( 'cache_dir' => APP_ROOT . '/src/Less/cache' );
+				$cssFileName = Less_Cache::Get( $lessFiles, $options );
+				$slim->response->headers->set( 'Content-Type', 'text/css' );
+				$slim->response->setBody( file_get_contents( APP_ROOT . '/src/Less/cache/' . $cssFileName ) );
```
MusikAnimal: This is where the magic happens. Requests to `index.css` will have to go through the stack, but the file is cached until the `.less` file is changed, so this should have noticeable affect on page load time.

We only to compile `index.less`. From there we can import Bootstrap or any other Less files, but saving that for another PR :)

I could not get a `/css/index.css` route to work, for whatever reason, figured `index.css` works just as well, especially since once we start using bootstrap.less we'd only have one .css file anyway.

src/Controllers/CopyPatrol.php
```diff
@@ -87,43 +87,7 @@ public function setWikiprojectDao( $wikiprojectDao ) {
...
-				if ( in_array( $_GET['filter'], $filterTypeKeys ) ) {
-					$filter = $_GET['filter'];
-				} else {
-					$this->flashNow( 'error', 'Invalid filter. Values must be one of: ' . join( $filterTypeKeys, ', ' ) );
-				}
-			}
-		}
-
-		$records = $this->dao->getPlagiarismRecords( 50, $filter, $filterUser );
+		$records = $this->getRecords();
```
MusikAnimal: Same code, just moved to it's own function.

src/Controllers/CopyPatrol.php
```diff
@@ -160,9 +124,58 @@ protected function handleGet() {
...
+			}
+		}
+
 		$this->view->set( 'filter', $filter );
 		$this->view->set( 'filterTypes', $filterTypes );
-		$this->render( 'index.html' );
+
+		// make this easier when working locally
+		$numRecords = $_SERVER['HTTP_HOST'] === 'localhost' ? 5 : 50;
+		return $this->dao->getPlagiarismRecords( $numRecords, $filter, $filterUser );
```
MusikAnimal: We had talked about how we prefer to change to a lower number of records when working locally, so I figure this would be a handy dandy way to automate it. I've forgotten to change the `5` back to `50` a number of times now.

.gitignore
```diff
@@ -1,8 +1,8 @@
 .DS_Store
-public_html/.DS_Store
 vendor/*
 .idea/*
 replica.my.cnf
 .env
 .lighttpd.conf
-
+src/Less/cache/*
+!.gitkeep
```
MusikAnimal: the `.gitkeep` file is solely so we can commit the `src/Less/cache` directory. Otherwise you will have to create the directory manually (because everything else in it is ignored), and the Less cache files won't be generated, and you won't have any CSS =P

src/App.php
```diff
@@ -198,6 +199,16 @@ function () use ( $slim ) {
 					$page( 'callback' );
 				} )->name( 'oauth_callback' );
 			} );
+		$slim->get( '/index.css',
+			function () use ( $slim ) {
+				// Compile LESS if need be, otherwise serve cached asset
+				// Cached files get automatically deleted if they are over a week old
+				$lessFiles = array( APP_ROOT . '/src/Less/index.less' => '/copypatrol/' );
```
Niharika29: What does this line do? I'm guessing the first part is the source path of the file but the second? And hardcoding it here is not a good idea. It didn't work for me, routing issues. And it'll break our test instance /plagiabot. We can go with a new env variable to handle this.

src/App.php
```diff
@@ -198,6 +199,16 @@ function () use ( $slim ) {
 					$page( 'callback' );
 				} )->name( 'oauth_callback' );
 			} );
+		$slim->get( '/index.css',
```
Niharika29: Since you're fetching '/index.css', it makes more sense for you to put it under group('/') - Line 157 and just fetch 'index.css'.

src/Controllers/CopyPatrol.php
Niharika29: Good idea! :D

PR #5, opened 2016-06-16
Improve HTML semantics, unobtrusive JS, notice for staging env

MusikAnimal:
Only functionality or noticeable change: Showing a banner if they are /plagiabot. Everything else is just semantics / JS unobtrusiveness.

The HTML semantics obviously are nothing important, just a minor improvement to make it easier to read.

### DIFF
--- a/public_html/js/index.js
+++ b/public_html/js/index.js
@@ -1,104 +1,120 @@
-/**
- * Save a review
- * @param id int ID of the record
- * @param val string Save value 'fixed' or 'false'
- */
-function saveState( id, val ) {
-	var buttonId, unusedButtonId, buttonClass, unusedButtonClass;
-
-	if ( val === 'fixed' ) {
-		buttonId = '#success' + id;
-		unusedButtonId = '#danger' + id;
-		buttonClass = 'success';
-		unusedButtonClass = 'danger';
-	} else if ( val === 'false' ) {
-		buttonId = '#danger' + id;
-		unusedButtonId = '#success' + id;
-		unusedButtonClass = 'success';
-		buttonClass = 'danger';
-	}
-	$( buttonId ).removeClass( 'btn-' + buttonClass ).addClass( 'btn-' + buttonClass + '-clicked' ).blur();
-	$( unusedButtonId ).removeClass( 'btn-' + unusedButtonClass ).addClass( 'btn-secondary' ).prop( 'disabled', 'disabled' ).blur();
-
-	$.ajax( {
-		url: 'review/add',
-		data: {
-			id: id,
-			val: val
-		},
-		dataType: 'json'
-	} ).done( function ( ret ) {
-		if ( ret.user ) {
-			$reviewerNode = $( '.status-div-reviewer-' + id );
-			$reviewerNode.find( '.reviewer-link' ).prop( 'href', ret.userpage ).text( ret.user );
-			$reviewerNode.find( '.reviewer-timestamp' ).text( ret.timestamp );
-			$reviewerNode.fadeIn( 'slow' );
-		} else if ( ret.error === 'Unauthorized' ) {
-			alert( 'You need to be logged in to be able to review.' );
-			$( buttonId ).addClass( 'btn-' + buttonClass ).removeClass( 'btn-' + buttonClass + '-clicked' ).blur();
-			$( unusedButtonId ).removeClass( 'btn-secondary' ).prop( 'disabled', false ).addClass( 'btn-' + unusedButtonClass );
-		} else {
-			alert( 'There was an error in connecting to database.' );
-			$( buttonId ).addClass( 'btn-' + buttonClass ).removeClass( 'btn-' + buttonClass + '-clicked' ).blur();
-			$( unusedButtonId ).removeClass( 'btn-secondary' ).prop( 'disabled', false ).addClass( 'btn-' + unusedButtonClass );
-		}
+$( document ).ready( function() {
+	/** Listeners */
+	$('body').tooltip({
+		selector: '[data-toggle="tooltip"]'
+	});
+	$( '.records' ).on( 'click', '.js-save-state', function () {
+		saveState( $( this ).data( 'id' ), $( this ).data( 'status' ) );
 	} );
-}
+	$( '.js-load-more' ).on( 'click', loadMoreResults );
+	$( '.js-compare-button' ).on( 'click', function () {
+		// pass the dataset of the element as an object to toggleComparePane
+		toggleComparePane.call(this, this.dataset);
+	});
 
-/**
- * Open the compare pane and do an AJAX request to Copyvios to fetch comparison data
- * @param id Ithenticate ID of record
- * @param index Index of the link in the copyvios list for the record
- * @param copyvio Copyvio URL
- * @param diffId Oldid of diff
- */
-function toggleComparePane( id, index, copyvio, diffId ) {
-	var compareDiv = '#comp' + id + '-' + index;
-	$( compareDiv ).slideToggle( 500 );
-	if ( !$( compareDiv ).hasClass( 'copyvios-fetched' ) ) {
+	/**
+	 * Save a review
+	 * @param id int ID of the record
+	 * @param val string Save value 'fixed' or 'false'
+	 */
+	function saveState( id, val ) {
+		var buttonId, unusedButtonId, buttonClass, unusedButtonClass;
+
+		if ( val === 'fixed' ) {
+			buttonId = '#success' + id;
+			unusedButtonId = '#danger' + id;
+			buttonClass = 'success';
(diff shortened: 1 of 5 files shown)


public_html/templates/base.html
```diff
@@ -98,21 +97,21 @@
...
 			{% endblock %}
-		</div>
+		</section>
 		<div class="load-more-div">
 			{% block loadmore %}
-				<button id="btn-load-more" class="btn btn-md btn-primary" onclick="loadMoreResults()">Load More</button>
+				<button id="btn-load-more" class="btn btn-md btn-primary js-load-more">Load More</button>
 			{% endblock %}
 		</div>
 	</section>
```
MusikAnimal: This whole file is just wrapping the JS in `$(document).ready` and adding the necessary listeners at the top

public_html/templates/index.html
```diff
@@ -7,116 +7,136 @@
 			<span>
 				<a class="reviewer-link" href="{{ row.reviewed_by_url }}" target="_blank">{{ row.status_user }}</a>
 			</span><br>
-			<span class="reviewer-timestamp">{{ row.review_timestamp }}</span>
+			<time datetime="{{ row.review_timestamp }}" class="reviewer-timestamp">{{ row.review_timestamp }}</time>
```
MusikAnimal: Using `<time>` when something is strictly displaying a time; the `datetime` attribute tells the browser this is a structured datetime string, so it can be used or transformed; e.g. browser extensions that convert times, adding something to your calendar (which obviously they won't do here)

public_html/templates/index.html
```diff
@@ -7,116 +7,136 @@
...
-						{% if row.editor %}
-							<a href='{{ row.editor }}'
-							   class={{ row.editor_page_dead ? 'text-danger' : '' }}>{{ row.editor }}</a><br>
-							<small>
-								<span><a href='{{ row.editor_talk }}'
-								         class='{{ row.editor_talk_dead ? 'text-danger' : '' }}'
-								         target="_blank">Talk</a> | </span>
-								<span><a href='{{ row.editor_contribs }}' target="_blank">Contributions</a> </span><br>
-								<span>Edit count: {{ row.editcount }}</span>
+			<article class="record">
```
MusikAnimal: `<article>` is for a "self-contained composition" on the page, where it is independent of the other articles on the page. This usually goes inside a `<section>`

src/Controllers/CopyPatrol.php
```diff
@@ -123,6 +123,11 @@ protected function handleGet( $lastId = 0 ) {
 			$records[$key]['wikiprojects'] = $cleanWikiprojects;
 		}
 		$this->view->set( 'records', $records );
+
+		// determine if we are on the staging environment, so we can show a banner in the view
+		$rootUri = $this->slim->request->getRootUri();
+		$this->view->set( 'staging', preg_match( '/\/plagiabot/', $rootUri ) );
```
MusikAnimal: I think this will work!

public_html/templates/base.html
```diff
@@ -98,21 +97,21 @@
...
-		<div class="record-container">
+		<section class="record-container">
 			{% block records %}
 			{% endblock %}
-		</div>
+		</section>
 		<div class="load-more-div">
 			{% block loadmore %}
-				<button id="btn-load-more" class="btn btn-md btn-primary" onclick="loadMoreResults()">Load More</button>
+				<button id="btn-load-more" class="btn btn-md btn-primary js-load-more">Load More</button>
```
MusikAnimal: Adding `js-` prefix to a class name to indicate to someone reading the HTML that there is JavaScript attached to this element

public_html/js/index.js
```diff
@@ -1,104 +1,113 @@
...
+				lastId: lastId
+			}
+		} ).done( function ( ret ) {
+			$( '#btn-load-more' ).text( 'Load More' ).removeClass( 'btn-loading' );
+			$newRecords = $( ret ).find( '.record-container' );
+			console.log( $newRecords );
+			$( '.record-container' ).append( $newRecords.html() );
+		} );
+	}
+} );
```
MusikAnimal: This whole file is just wrapping the JS in `$(document).ready` and adding the necessary listeners at the top

src/Controllers/CopyPatrol.php
Niharika29: Let's move this to a middleware instead. This should not be a part of CopyPatrol controller.

src/App.php
```diff
@@ -161,6 +161,11 @@ function () use ( $slim ) {
 					$page->setDao( $slim->plagiabotDao );
 					$page->setEnwikiDao( $slim->enwikiDao );
 					$page->setWikiprojectDao( $slim->wikiprojectDao );
+
+					// determine if we are on the staging environment, so we can show a banner in the view
+					$rootUri = $slim->request->getRootUri();
+					$slim->view->set( 'staging', preg_match( '/\/plagiabot/', $rootUri ) );
```
MusikAnimal: @Niharika29 I assume this is where you meant? Tried it out and it worked

Niharika29: Not exactly, but I'll take care of this now and merge in a bit. Because we might need to do this in several places, it makes more sense to add it to the middleware array above (L143) and then simply include it, like I've done with inject-user.

public_html/templates/base.html
```diff
@@ -38,16 +33,20 @@
...
-				<a href="https://meta.wikimedia.org/wiki/Talk:Community_Tech/Improve_the_plagiarism_detection_bot">
-					You can drop your feedback here</a>. <a href="https://tools.wmflabs.org/copypatrol/">Copypatrol</a>
-				is our live, ready-to-use
-				tool, while <a href="https://tools.wmflabs.org/plagiabot/">Plagiabot</a> is our testing playground.
-				Please do not use Plagiabot to record your edits.
-			</div>
+			{% if staging %}
+				<div class="alert alert-danger text-center">
+					This is a staging environment. For the actual CopyPatrol tool, please use
+					<a href="/copypatrol">tools.wmflabs.org/copypatrol</a>
```
Niharika29: Changed the href to complete url. PHPStorm keeps throwing errors otherwise.

public_html/templates/index.html
```diff
@@ -7,116 +7,147 @@
...
+				<div class="compare-links-container">
+					<a class="btn btn-xs btn-primary compare-button" href='{{ row.turnitin_report }}' target="_blank">
+						<span class="glyphicon glyphicon-new-window"></span>
+						Turnitin report
+					</a><br>
+					{% if row.copyvios is not empty %}
+						{% for copy in row.copyvios %}
+							<div class="compare-div">
+								<button class="btn btn-xs btn-primary dropdown-toggle compare-button"
+												onclick="toggleComparePane( {{ row.ithenticate_id }}, {{ loop.index }}, '{{ copy }}', {{ row.diff }} )">
```
Niharika29: I am guessing you didn't change this the way you changed others because it's trickier? Or did you just miss it? :)

src/App.php
kaldari: Not that it makes much difference since it's just a URL, but strpos() is better to use in this case than preg_match() as it's more efficient (it doesn't use the regex engine).

public_html/js/index.js
```diff
@@ -1,104 +1,120 @@
...
+	$('body').tooltip({
+		selector: '[data-toggle="tooltip"]'
+	});
+	$( '.records' ).on( 'click', '.js-save-state', function () {
+		saveState( $( this ).data( 'id' ), $( this ).data( 'status' ) );
+	} );
+	$( '.js-load-more' ).on( 'click', loadMoreResults );
+	$( '.js-compare-button' ).on( 'click', function () {
+		// pass the dataset of the element as an object to toggleComparePane
+		toggleComparePane.call(this, this.dataset);
```
MusikAnimal: @Niharika29 I did forget to do `toggleComparePane`, here it is. Since there are so many values we need it seems easier to just pass it in as a key/value hash

public_html/templates/base.html
MusikAnimal: Interesting. I don't think there's anything invalid with using a relative path, and in this case it'd be necessary for the link to work on both our local and on Tool Labs – not that we need it to work on our local

public_html/js/index.js
Niharika29: I merged the branch without this last patch, for toggleComparePane because it doesn't bind to dynamically created records (with load more functionality). I researched around and found this: http://stackoverflow.com/questions/203198/event-binding-on-dynamically-created-elements so we should be binding to the parent and not the dynamically created child. This would be problematic because we can't find individual compare records without the ithenticate ID and loop index.

Niharika29:
Merged without the last bit.

src/App.php
Niharika29: Done.

PR #7, opened 2016-06-17
Show 'all reviewed cases' instead of fixed or no action

MusikAnimal:
Bug: T132352

### DIFF
--- a/src/Dao/PlagiabotDao.php
+++ b/src/Dao/PlagiabotDao.php
@@ -48,8 +48,10 @@ class PlagiabotDao extends AbstractDao {
 
 	/**
 	 * @param int $n Number of records asked for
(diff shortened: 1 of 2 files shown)


src/Dao/PlagiabotDao.php
```diff
@@ -48,7 +48,7 @@ public function __construct( $dsn, $user, $pass, $wiki = 'https://en.wikipedia.o
 
 	/**
 	 * @param int $n Number of records asked for
-	 * @param string $filter Filter SQL to show a certian status, one of 'all', 'fixed', 'noaction', 'open' or 'mine'
+	 * @param string $filter Filter SQL to show a certian status, one of 'all', 'open', 'reviewed' or 'mine'
 	 * @param string $filterUser Filter SQL to only return records reviewed by given user
```
Niharika29: This PHPDoc block should be updated now that we're passing options array instead of filter and filterUser (sorry, I should have done this in my fix.)

PR #8, opened 2016-06-17
Fix timestamp formatting

MusikAnimal:


src/Controllers/CopyPatrol.php
```diff
@@ -208,11 +208,16 @@ public function getReportLink( $ithenticateId ) {
...
 	 * @return string Reformatted date
 	 */
-	public function formatTimestamp( $datetime ) {
+	public function formatTimestamp( $datetime, $gmt = false ) {
 		$datetime = strtotime( $datetime );
-		return date( 'Y-m-d h:m', $datetime );
+		if ( $gmt ) {
+			return gmdate( 'Y-m-d H:i', $datetime );
+		} else {
+			return date( 'Y-m-d H:i', $datetime );
```
bd808: These two invocations will be the same from Tool Labs where the local server timezone is UTC. It really might be easier/better to change https://github.com/Niharika29/PlagiabotWeb/blob/master/public_html/index.php#L33-L38 so that the app always sets the timezone to UTC.

bd808:
LGTM

src/Controllers/Review.php
```diff
@@ -47,7 +47,7 @@ protected function handleGet() {
 				array(
 					'user' => $user,
 					'userpage' => $this->getUserPage( $user ),
-					'timestamp' => $this->formatTimestamp( $timestamp )
+					'timestamp' => $this->formatTimestamp( $timestamp, true ) // 'true' to convert to GMT
```
ricordisamoa: https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP#Functions_and_parameters

MusikAnimal:
Moving to dedicated branch, see #12

PR #9, opened 2016-06-20
Add ability to undo your reviews

MusikAnimal:
Phab: https://phabricator.wikimedia.org/T134597

This PR also includes code from https://github.com/Niharika29/PlagiabotWeb/pull/8

For the purposes of reviewing, **ignore all .less files** except `index.less`. 72 of the 78 files changed in this PR are just the Bootstrap Less files being put into version control. Unfortunately there is no composer package or other graceful way of importing these Less files :/


MusikAnimal:
Moving to dedicated branch, see #12

PR #10, opened 2016-06-20
Compare labels and history link

MusikAnimal:
Not as much code as it looks like; this includes the changes in #8 and #9 (there are some dependencies)


MusikAnimal:
Moving everything to dedicated branch, see #12

PR #12, opened 2016-06-21
Wikimania branch

MusikAnimal:
All of Leon's stuff during Wikimania. Each commit represents a specific task, except "Add ability to undo your reviews" which includes the fix for [T138098](https://phabricator.wikimedia.org/T138098), see https://github.com/Niharika29/PlagiabotWeb/commit/1062088737d5a4933779b2d850f9ecebdf5a42c8#diff-0a3a612b6e041fccdbde3eaa6b795886R215

For the purposes of reviewing, **ignore all .less files** except `index.less`. 72 of the 80 files changed in this PR are just the Bootstrap Less files being put into version control. Unfortunately there is no composer package or other graceful way of importing these Less files :/

### DIFF
--- a/public_html/js/index.js
+++ b/public_html/js/index.js
@@ -4,12 +4,20 @@ $( document ).ready( function () {
 		selector: '[data-toggle="tooltip"]'
 	} );
 	$( '.records' ).on( 'click', '.js-save-state', function () {
-		saveState( $( this ).data( 'id' ), $( this ).data( 'status' ) );
+		var status = $( this ).data( 'status' ),
+			id = $( this ).data( 'id' );
+
+		// undo review if they click on the button with the same status as the record
+		if ( status === $( '.record-' + id ).data( 'status' ) ) {
+			undoReview( id, status );
+		} else {
+			saveState( id, status );
+		}
 	} );
 	$( '.records' ).on( 'click', '.js-compare-button', function () {
 		// pass the dataset of the element as an object to toggleComparePane
 		toggleComparePane.call(this, this.dataset);
-	});
+	} );
 	$( '.js-load-more' ).on( 'click', loadMoreResults );
 
 	/**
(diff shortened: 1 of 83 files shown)


public_html/js/index.js
```diff
@@ -4,12 +4,20 @@ $( document ).ready( function () {
...
+			undoReview( id, status );
+		} else {
+			saveState( id, status );
+		}
 	} );
 	$( '.records' ).on( 'click', '.js-compare-button', function () {
 		// pass the dataset of the element as an object to toggleComparePane
 		toggleComparePane.call(this, this.dataset);
-	});
+	} );
```
Niharika29: }) is the standard when doing MW development. I didn't find anything in the Manual for coding conventions but you'll find examples in the code. I just rely on Bryan's code to check myself: https://github.com/wikimedia/wikimedia-iegreview/blob/master/public/js/site.js#L65

public_html/js/index.js
```diff
@@ -49,17 +44,61 @@ $( document ).ready( function () {
...
+	 * Undo a review
+	 * @param id int ID of the record
+	 * @param oldStatus string current review state of the record
+	 */
+	function undoReview( id, oldStatus ) {
+		// update styles before AJAX to make it seem more responsive
+		setReviewState( id, 'open' );
+
+		$.ajax( {
+			url: 'review/add',
```
Niharika29: I think we should do the Undo part with a new route 'review/undo' and write a ReviewUndo controller for it. We can move all the Review controllers in a Review directory. It seems far neater.

src/Controllers/Review.php
```diff
@@ -35,19 +35,27 @@ public function __construct( \Slim\Slim $slim = null ) {
 
 	protected function handleGet() {
 		$id = $this->request->get( 'id' );
-		$val = $this->request->get( 'val' );
+		$undo = (bool)$this->request->get( 'undo' );
```
Niharika29: See my comment above about having this in a separate controller.

src/Less/index.less
```diff
@@ -208,11 +225,11 @@ a[href=''] {
 }
 
 .cv-hl-in {
-  background: linear-gradient(to left, #FFC264, white);
+  #gradient.horizontal(white, #FFC264);
```
Niharika29: Is there a benefit to using one over the other?

public_html/js/index.js
MusikAnimal: It seems Bryan wasn't going after conventions there; there's no [whitespace](https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Whitespace) at all

src/Less/index.less
MusikAnimal: Mostly just a nifty shorthand, but it does also add vendor prefixing: https://github.com/Niharika29/PlagiabotWeb/pull/12/files#diff-f7cc4de34a9083a20c1c95fd3725b03eR9

src/App.php
```diff
@@ -197,17 +197,25 @@ function () use ( $slim ) {
...
-					// but I haven't found a way to make it work with AJAX requests so far
-					if ( $slim->authManager->isAuthenticated() ) {
-						$page = new Controllers\Review( $slim );
-						$page->setDao( $slim->plagiabotDao );
-						$page();
-					} else {
+				// TODO: Ideally the authentication step should be taken care of by a middleware
+				// but I haven't found a way to make it work with AJAX requests so far
+				$requireAuth = function () use ( $slim ) {
+					if ( !$slim->authManager->isAuthenticated() ) {
```
Niharika29: You should move this function to the middleware array above. Line 143.

Niharika29: And also just move it to Line 200.

PR #13, opened 2016-06-22
Make batch requests/queries to API and database to improve performance

MusikAnimal:
From my testing, this improves load time dramatically. Everything but querying for WikiProjects is done en masse. The biggest slowdown was the API, so checking for dead links in only three different requests makes a big difference, it seems.

This PR I'm not super confident in, but it works :) I know I need to use prepared statements in `getRevisionDetailsMulti`, but I could not get it to work for some reason. Meanwhile `handleGet` looks very messy, but this sequence of loops is necessary to ensure we have all the data we need before updating the view – and doing so in a way that is performant.

### DIFF
--- a/src/Controllers/CopyPatrol.php
+++ b/src/Controllers/CopyPatrol.php
@@ -89,23 +89,60 @@ class CopyPatrol extends Controller {
 	 */
 	protected function handleGet() {
 		$records = $this->getRecords();
+		$diffIds = array();
+		$pageTitles = array();
+		$editors = array();
+		$editorPages = array();
+		$editorTalkPages = array();
+
+		// first build arrays of diff IDs and page titles so we can use them to make mass queries
+		foreach ( $records as $record ) {
+			$diffIds[] = $record['diff'];
+			$pageTitles[] = $record['page_title'];
+		}
+
+		// get info for each revision (editor, editcount, etc) and build datasets from it
+		$revisions = $this->enwikiDao->getRevisionDetailsMulti( $diffIds );
+		foreach ( $revisions as $revision ) {
+			$userText = $revision['rev_user_text'];
+			if ( isset( $userText ) ) {
+				// associative array for editor info with the revision ID as the key,
+				// this makes it easier to access what we need when looping through the copyvio records
+				$editors[$revision['rev_id']] = array(
+					'editor' => $userText,
+					'editcount' => $revision['user_editcount']
+				);
+				// build arrays for editor page and talk page so we can mass-query if they are dead
+				$editorPages[] = 'User:' . $userText;
+				$editorTalkPages[] = 'User talk:' . $userText;
+			}
+		}
+		// get all the dead pages in 3 goes; these cannot be done at the same time as we can only query for 50 pages max
+		$deadPages = $this->enwikiDao->getDeadPages( $pageTitles );
+		$deadEditorPages = $this->enwikiDao->getDeadPages( $editorPages );
+		$deadEditorTalkPages = $this->enwikiDao->getDeadPages( $editorTalkPages );
+
+		// now all external requests and database queries (except WikiProjects) have been completed,
+		// let's loop through the records once more to build the complete dataset to be rendered into view
 		foreach ( $records as $key => $record ) {
-			$editor = $this->enwikiDao->getUserDetails( $record['diff'] );
+			$editor = isset( $editors[$record['diff']] ) ? $editors[$record['diff']] : NULL;
 			$records[$key]['diff_timestamp'] = $this->formatTimestamp( $record['diff_timestamp'] );
 			$records[$key]['diff_link'] = $this->getDiffLink( $record['page_title'], $record['diff'] );
 			$records[$key]['page_link'] = $this->getPageLink( $record['page_title'] );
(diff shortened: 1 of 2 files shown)


src/Dao/EnwikiDao.php
```diff
@@ -47,6 +47,29 @@ public function __construct( $dsn, $user, $pass, $wiki = 'https://en.wikipedia.o
...
+	 *
+	 * @param $diffs array Revision IDs
+	 * @return array full revision rev_id, rev_user, rev_usertext, user_editcount and user_name of the revision
+	 */
+	public function getRevisionDetailsMulti( $diffs ) {
+		$query = self::concat(
+			'SELECT r.rev_id, r.rev_user, r.rev_user_text, u.user_editcount, u.user_name',
+			'FROM revision r',
+			'LEFT JOIN user u ON r.rev_user = u.user_id',
+			'WHERE r.rev_id IN (' . join( ',', $diffs ) . ')' // FIXME: could not get prepared statement (?) to work
```
bd808: To do this as a prepared statement you will need to insert a placeholder for each value in $diffs. Something like:

```
'WHERE r.rev_id IN (' . implode( ',', array_fill( 0, count( $diffs ), '?' ) ) . ')'
```
Then you can pass `$diffs` directly as the parameter collection to `fetchAll()`.

src/Controllers/CopyPatrol.php
```diff
@@ -89,22 +89,59 @@ public function setWikiprojectDao( $wikiprojectDao ) {
...
+
+		// first build arrays of diff IDs and page titles so we can use them to make mass queries
+		foreach ($records as $record) {
+			$diffIds[] = $record['diff'];
+			$pageTitles[] = $record['page_title'];
+		}
+
+		// get info for each revision (editor, editcount, etc) and build datasets from it
+		$revisions = $this->enwikiDao->getRevisionDetailsMulti( $diffIds );
+		foreach ($revisions as $revision) {
```
Niharika29: foreach ( $revisions as $revision ) { 
}

Line 99 as well.

src/Dao/EnwikiDao.php
```diff
@@ -75,6 +98,37 @@ public function getUserDetails( $diff ) {
...
+		curl_setopt( $ch, CURLOPT_URL, $url );
+		curl_setopt( $ch, CURLOPT_RETURNTRANSFER, TRUE );
+		$result = curl_exec( $ch );
+		$json = json_decode( $result );
+		$deadPages = array();
+		foreach ( $json->query->pages as $p ) {
+			if ( isset( $p->missing ) ) {
+				// Please note that this returns a false positive when the user account has a global User page and
+				// not a local one
+				array_push( $deadPages, $p->title );
```
Niharika29: $deadPages[] = $p->title;

src/Controllers/CopyPatrol.php
```diff
@@ -164,7 +201,7 @@ protected function getRecords() {
 			}
 		}
 		// make this easier when working locally
-		$numRecords = $_SERVER['HTTP_HOST'] === 'localhost' ? 3 : 50;
+		$numRecords = $_SERVER['HTTP_HOST'] === 'localhost' ? 10 : 50;
```
bd808: Not for this PR, but this limit should probably be a configuration option rather than relying on inspecting HTTP_HOST.

src/Dao/EnwikiDao.php
```diff
@@ -47,6 +47,28 @@ public function __construct( $dsn, $user, $pass, $wiki = 'https://en.wikipedia.o
...
+	 * @return array full revision rev_id, rev_user, rev_usertext, user_editcount and user_name of the revision
+	 */
+	public function getRevisionDetailsMulti( $diffs ) {
+		$query = self::concat(
+			'SELECT r.rev_id, r.rev_user, r.rev_user_text, u.user_editcount, u.user_name',
+			'FROM revision r',
+			'LEFT JOIN user u ON r.rev_user = u.user_id',
+			'WHERE r.rev_id IN (' . implode( ',', array_fill( 0, count( $diffs ), '?' ) ) . ')'
+		);
+		$data = array(
```
bd808: Unused?

src/Dao/EnwikiDao.php
```diff
@@ -75,6 +97,37 @@ public function getUserDetails( $diff ) {
...
 
 	/**
+	 * Determine which of the given pages are dead
+	 *
+	 * @param $titles array Page titles
+	 * @return array the pages that are dead
+	 */
+	public function getDeadPages( $titles ) {
+		if ( !$titles ) {
+			return false;
```
bd808: Should probably either return an empty array or you should change the phpdoc for the return value to `@return array|bool`

src/Dao/EnwikiDao.php
```diff
@@ -75,6 +97,37 @@ public function getUserDetails( $diff ) {
...
+		}
+		$titles = array_map( function( $title ) {
```
bd808: You don't need the closure here, just use `$titles = array_map( 'urlencode', $title );`

bd808:
LGTM

MusikAnimal:
Alrighty, going to take the liberty of merging and deploying :) Thanks for the help, all!

PR #14, opened 2016-06-23
Trim trailing slash from routes, if present

MusikAnimal:
Bug: T138475

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -36,23 +36,22 @@ class App extends AbstractApp {
 	 * @param \Slim\Slim $slim Application
 	 */
 	protected function configureSlim( \Slim\Slim $slim ) {
-		$slim->config(
-			[ 'displayErrorDetails' => true,
-			  'debug' => true,
-			  'oauth.enable' => Config::getBool( 'USE_OAUTH', false ),
-			  'oauth.consumer_token' => Config::getStr( 'OAUTH_CONSUMER_TOKEN' ),
-			  'oauth.secret_token' => Config::getStr( 'OAUTH_SECRET_TOKEN' ),
-			  'oauth.endpoint' => Config::getStr( 'OAUTH_ENDPOINT' ),
-			  'oauth.redir' => Config::getStr( 'OAUTH_REDIR' ),
-			  'oauth.callback' => Config::getStr( 'OAUTH_CALLBACK' ),
-			  'db.dsnwp' => Config::getStr( 'DB_DSN_WIKIPROJECT' ),
-			  'db.dsnen' => Config::getStr( 'DB_DSN_ENWIKI' ),
-			  'db.dsnpl' => Config::getStr( 'DB_DSN_PLAGIABOT' ),
-			  'db.user' => Config::getStr( 'DB_USER' ),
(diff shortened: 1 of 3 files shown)


src/App.php
```diff
@@ -36,23 +36,22 @@ class App extends AbstractApp {
...
+			'oauth.endpoint' => Config::getStr( 'OAUTH_ENDPOINT' ),
+			'oauth.redir' => Config::getStr( 'OAUTH_REDIR' ),
+			'oauth.callback' => Config::getStr( 'OAUTH_CALLBACK' ),
+			'db.dsnwp' => Config::getStr( 'DB_DSN_WIKIPROJECT' ),
+			'db.dsnen' => Config::getStr( 'DB_DSN_ENWIKI' ),
+			'db.dsnpl' => Config::getStr( 'DB_DSN_PLAGIABOT' ),
+			'db.user' => Config::getStr( 'DB_USER' ),
+			'db.pass' => Config::getStr( 'DB_PASS' ),
+			'templates.path' => '../public_html/templates'
+		] );
```
MusikAnimal: matching indentation / bracket style with how we do our functions, hopefully this is OK

src/App.php
```diff
@@ -163,11 +162,18 @@ protected function configureRoutes( \Slim\Slim $slim ) {
 					echo json_encode( array( 'error' => 'Unauthorized' ) );
 					$slim->stop();
 				}
+			},
+			'trailing-slash' => function () use ( $slim ) {
+				// Remove trailing slashes
+				if ( substr( $_SERVER['REQUEST_URI'], -1 ) === '/' ) {
+					$uri = preg_replace( '/\/+$/', '', $_SERVER['REQUEST_URI'] );
```
bd808: Or `$uri = rtrim( $_SERVER['REQUEST_URI'], '/' );`.

MusikAnimal: that's what I had originally, but it will remove only one `/` then redirect, so Chrome will error out `tools.wmflabs.org/plagiabot/////////` with too many redirects

src/App.php
```diff
@@ -163,11 +162,18 @@ protected function configureRoutes( \Slim\Slim $slim ) {
...
+					$slim->redirect( $uri );
+				}
 			}
 		);
-		$slim->group( '/', $middleware['inject-user'], $middleware['set-environment'],
+		$slim->group( '/', $middleware['trailing-slash'], $middleware['inject-user'], $middleware['set-environment'],
 			function () use ( $slim ) {
-				$slim->get( '/?', function () use ( $slim ) {
+				$slim->get( '/', function () use ( $slim ) {
```
bd808: Shouldn't this route be for `''` now that you are ensuring that there are no trailing slashes? Since you are in the `'/'` group this seems like it should only match an URL of `/tool-name//`.

src/App.php
MusikAnimal: oh wait I see

src/App.php
MusikAnimal: looks like we need it to get `index.css` to work

MusikAnimal:
Merging... I keep going to /copypatrol (without trailing slash) and ze buttons don't work. Seems like a high-priority fix :-/ We can refine the logic as needed

PR #16, opened 2016-07-06
Utilize addwiki API framework, replacing labs db queries with API queries

MusikAnimal:


src/Controllers/CopyPatrol.php
```diff
@@ -22,6 +22,8 @@
 namespace Plagiabot\Web\Controllers;
 
 use Wikimedia\Slimapp\Controller;
+use GuzzleHttp;
+use GuzzleHttp\Promise\Promise;
```
MusikAnimal: Note to self: Don't ever use cURL again. Guzzle it down!

src/Dao/EnwikiDao.php
```diff
@@ -44,26 +48,71 @@ public function __construct(
...
+			}
+		}
+
+		foreach ( $result['pages'] as $page ) {
+			$revisions = $page['revisions'];
+
+			if ( isset( $revisions ) ) {
+				foreach ( $revisions as $revision ) {
+					$data[$revision['revid']] = [
+						'revid' => $revision['revid'],
```
kaldari: It's a bit odd having the revision ID in the data twice. Do we need both of them?

MusikAnimal: Nope. We actually don't need the timestamp either... I can remove both and rename the function to `getRevisionEditors` or something

kaldari: That sounds like a good idea.

PR #18, opened 2016-07-10
Add ORES scores

Ladsgroup:
I have tested functions I added but I couldn't test it in local app. 

Bug: T139009

### DIFF
--- a/public_html/templates/index.html
+++ b/public_html/templates/index.html
@@ -69,6 +69,11 @@
 							{% endfor %}
 						{% endif %}
 					</div>
+					<div class="row-div col-sm-2 text-center ores-div">
+						{% if row.oresScore is not empty %}
+							{{ row.oresScore }}
+						{% endif %}
+					</div>
 					<div class="row-div col-sm-2 text-center status-div">
 						<button data-id="{{ row.ithenticate_id }}"
 						        data-status="fixed"
--- a/public_html/templates/index.html
+++ b/public_html/templates/index.html
@@ -38,7 +38,7 @@
 						{% if row.oresScore is not empty %}
 						<br/>
 						<small class="diff-div-ores">
-							ORES score: {{ row.oresScore }}
+							ORES score: {{ row.oresScore }}%
 						</small>
 						{% endif %}
 					</div>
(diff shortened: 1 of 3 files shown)


kaldari:
I'm not sure it really makes sense to have a whole column devoted to the ORES score, especially since ORES isn't tuned to detect plagiarism yet. I think it would make more sense to list it in the Diff column, under the timestamp, and perhaps only if the score is significant.

Niharika29:
@Ladsgroup thank you for this patch!
I will merge it after the Phabricator ticket gets discussed in our team meeting. 
We will also need some way to let people know what the ORES score actually is but that can come in a separate patch and will need some design input.

Ladsgroup:
Hey Niharika! Thank you. I implemented Ryan's idea and put ORES scores under timestamp. It would be great!

MusikAnimal:
@Ladsgroup sorry, I merged another branch and I guess it conflicts with this one. Should be easy to resolve, let me know if you need help!

Ladsgroup:
I fixed the conflicts :)

kaldari:
@Ladsgroup: We discussed this patch at our sprint planning meeting today and we decided that we only want to show the ORES score if it's a very high score, i.e. likely to be vandalism. The reason for this is that we are also going to be showing the Turnitin scores (https://phabricator.wikimedia.org/T139209), and we don't want to confuse people with a lot of different scores unless it's likely to be important information.

public_html/templates/index.html
```diff
@@ -35,6 +35,12 @@
 						<small class="diff-div-timestamp">
 							<time datetime="{{ row.diff_timestamp }}">{{ row.diff_timestamp }}</time>
 						</small>
+						{% if row.oresScore is not empty %}
+						<br/>
+						<small class="diff-div-ores">
+							ORES score: {{ row.oresScore }}
```
kaldari: ... and move the '%' here.

public_html/templates/index.html
```diff
@@ -35,6 +35,12 @@
 						<small class="diff-div-timestamp">
 							<time datetime="{{ row.diff_timestamp }}">{{ row.diff_timestamp }}</time>
 						</small>
+						{% if row.oresScore is not empty %}
```
kaldari: Let's put a threshold here so that we're only showing it for high scores...

Ladsgroup:
Done

PR #19, opened 2016-07-11
Add leaderboard page showing top reviewers

MusikAnimal:


public_html/templates/base.html
```diff
@@ -19,9 +19,10 @@
 </head>
 <body>
 <header class="app-header">
-	<img src="{{ siteUrl( 'images/copypatrol.png' ) }}" />
+	<a href="{{ siteUrl( '' ) }}"><img src="{{ siteUrl( 'images/copypatrol.png' ) }}" /></a>
```
kaldari: Good idea.

src/Controllers/Leaderboard.php
```diff
@@ -0,0 +1,46 @@
...
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ * more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author MusikAnimal <musikanimal@wikimedia.org>
+ * @copyright © 2016 Niharika Kohli and contributors.
```
kaldari: At some point we need to change all these copyright notices to say "@copyright © 2016 Wikimedia Foundation", but that's not really relevant to this change.

public_html/templates/base.html
```diff
@@ -73,43 +74,46 @@
...
-	</section>
+	{% block main %}
+		<section class="filters container-fluid">
+			<form>
+				<fieldset class="form-group filters-fieldset text-center">
+					{% for type, description in filterTypes %}
+						<span class="radio-inline">
+							<label>
+								<input type="radio" name="filter" value="{{ type }}"
+										{% if filter == type %} checked="checked"{% endif %}
```
kaldari: Extra tab.

public_html/templates/leaderboard.html
```diff
@@ -0,0 +1,45 @@
...
+{% block main %}
+  <section class="leaderboard container">
+    <a class="back-to-copypatrol" href="{{ siteUrl( '' ) }}">
+      <span class="glyphicon glyphicon-chevron-left"></span>
+      Back to CopyPatrol
+    </a>
+    <h1 class="text-center">Leaderboard</h1>
+
+    <div class="text-center clearfix">
+      <span class="leaderboard-column col-sm-4">
```
Niharika29: Why are these spans and not div? Aren't spans usually meant for inline things only?

MusikAnimal: They are, looks like the browser is converting them to `display: block` when they are side by side and have a `float` property, which is an interesting find. Semantically and visually they are still inline, however, and they actually become inline when the columns are broke on to separate lines – which is when I would expect it be `display:block`, not the other way around!

PR #21, opened 2016-07-18
Add WikiProject filtering with Select2 and the WP bubble links

MusikAnimal:
A few notes:
* WikiProjects are submitted to the server in the format `Medicine|Military History|Something` mainly to be Select2-friendly, since setting the values of the Select2 control is super buggy, so we instead use a hidden input field. Additionally this follows the standard pipe-separated titles format we see in MediaWiki
* Despite my initial testing, this still runs a little slow. We have to use regex to query for multiple WikiProjects


public_html/templates/base.html
```diff
@@ -75,25 +77,39 @@
 
 	{% block main %}
 		<section class="filters container-fluid">
-			<form>
-				<fieldset class="form-group filters-fieldset text-center">
-					{% for type, description in filterTypes %}
-						<span class="radio-inline">
+			<form id="filters-form">
+				<fieldset class="form-group filters-fieldset col-lg-6">
```
Niharika29: col-sm-6. We're using sm everywhere else to present it better on small screens. (in other places too)

Niharika29: Hmm, I'd personally prefer the filters centered. Looks a bit strange currently.

MusikAnimal: Here I don't think `lg` vs `sm` makes a difference as there are no columns, just two rows, which begs the question if we even need gird classes. I agree the left-alignment looks weird but I tried centering and it didn't look too great either. Here's a screenshot:
screen shot 2016-07-21 at 12 50 13 am
If you're cool with that I am too, maybe rope in Danny on the decision.

public_html/templates/index.html
```diff
@@ -71,7 +71,7 @@
 					<div class="row-div col-sm-3 text-center wikiproject-div">
 						{% if row.wikiprojects is not empty %}
 							{% for wp in row.wikiprojects %}
-								<span class="row-div wproject">{{ wp }}</span>
+								<a class="row-div wproject" href="#">{{ wp }}</a>
```
Niharika29: This messes up the readability of wikiprojects a bit. Let's make the bubble background lighter. #f3f2f2 or something.

public_html/templates/base.html
Niharika29: Centered looks much better! 
Here's what I see on my phone:
img_8424
I thought this was because of lg instead of sm but might not be so.

public_html/templates/index.html
MusikAnimal: We could also make the links black but keep their hover state, or even have the whole bubble change to a different colour on hover. Light gray background works too, I think I've just become so accustomed to `#d0d0d0` :)

Niharika29: Changing on hover won't work since we're trying to improve readability, people shouldn't have to hover for that. I leave the decision for this up to you. :)

public_html/templates/base.html
MusikAnimal: Your phone is huge!!!! :) I guess there actually is a grid class on the select box, but I don't think we need them at all. I'll push back some updates shortly, and also make them centered.

public_html/templates/index.html
MusikAnimal: No I meant like make the bubbles act like buttons. We could then keep the black and gray as we have it now. I suppose many won't realize they're clickable until you hover over it, though

public_html/templates/base.html
Niharika29: Heh, it isn't, GitHub insists on stretching it for some reason. :/ 
I don't think it hurts to keep the grid classes because we're using them everywhere but you're the CSS guru. :P 
Thanks! :)

public_html/templates/index.html
Niharika29: When I was working on this app in the very early stages I thought I'd color each Wikiproject bubble uniquely, so every project has an associated color and that's the color you see in the SelectJS too. I slowly realized this was not worth the amount of work behind it though.

MusikAnimal: We should look into it! Ryan already [created a palette](https://github.com/MusikAnimal/pageviews/issues/25#issuecomment-186755073) of 10 colours that are colour-blind friendly, and we have the Select2 code for it. I think it'd make the tool look really fun if the WikiProjects column looked like Fruit Loops cereal

kaldari: Yeah, I don't think it's worth separately color-coding each WikiProject. There are about 1000 different WikiProjects so I'm not going to remember what project is what color, and even if I do, it will be roughly the same color as 100 other WikiProjects.

MusikAnimal: Well I figure we'd rotate between the 10, starting from the top of the page. I guess there's the complication that if they select a WikiProject via search, it's colour would need to match that of the same WikiProject bubble on the page. But yes, this effort is totally unnecessary and not worth our while =P

src/Dao/PlagiabotDao.php
```diff
@@ -76,29 +85,44 @@ public function getPlagiarismRecords( $n = 50, $options ) {
...
+			$wikiprojectsSql = self::concat(
+				'INNER JOIN wikiprojects',
+				'ON wp_page_title = page_title'
+			);
+			$filters[] = 'wp_project RLIKE :wikiprojects';
+		}
+
 		// construct necessary SQL based on filters
 		if ( !empty( $filters ) ) {
 			$filterSql = 'WHERE ' . join( ' AND ', $filters );
```
bd808: `self::buildWhere( $filters )`

bd808:
👍 for the use of `IN` instead of `RLIKE`.

public_html/js/index.js
```diff
@@ -185,5 +194,57 @@
...
+					url: 'https://en.wikipedia.org/w/api.php',
+					dataType: 'jsonp',
+					delay: 200,
+					jsonpCallback: $.noop(), // have to provide some function or else Select2 gets angry
+					data: function( search ) {
+						return {
+							action: 'query',
+							list: 'prefixsearch',
+							format: 'json',
+							pssearch: 'Wikipedia:WikiProject ' + ( search.term || '' )
```
kaldari: We should be excluding redirects here, since some redirects that look like WikiProjects aren't, for example:
https://en.wikipedia.org/w/index.php?title=Wikipedia:WikiProject_Animals_in_The_Signpost&redirect=no
https://en.wikipedia.org/w/index.php?title=Wikipedia:WikiProject_Comics_in_The_Signpost&redirect=no

MusikAnimal: Good call!

MusikAnimal: @kaldari fixed with https://github.com/Niharika29/PlagiabotWeb/commit/fe24c106a557eddb84b73a72a38e92c91e47bb8d

PR #22, opened 2016-07-26
Show only records after June 20, 2016

MusikAnimal:
Bug: T138317


src/Dao/PlagiabotDao.php
```diff
@@ -113,6 +113,9 @@ public function getPlagiarismRecords( $n = 50, $options ) {
 			$filters[] = "wp_project IN ($bindParams)";
 		}
 
+		// show only records after June 20, 2016; See phab:T138317
+		$filters[] = "diff_timestamp > '20160620000000'";
```
kaldari: Seems strange to put quotation marks around the number here.

kaldari:
+2

PR #23, opened 2016-07-28
Include scores for each source URL

MusikAnimal:
Bug: T139209

### DIFF
--- a/src/Controllers/CopyPatrol.php
+++ b/src/Controllers/CopyPatrol.php
@@ -162,7 +162,7 @@ class CopyPatrol extends Controller {
 			$records[$key]['page_link'] = $this->getPageLink( $record['page_title'] );
 			$records[$key]['history_link'] = $this->getHistoryLink( $record['page_title'] );
 			$records[$key]['turnitin_report'] = $this->getReportLink( $record['ithenticate_id'] );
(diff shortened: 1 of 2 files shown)


src/Controllers/CopyPatrol.php
```diff
@@ -426,30 +426,29 @@ public function getUserContribs( $user ) {
...
-						$isSubstring = true;
-					}
-				}
-				if ( $isSubstring === false ) {
-					$uniqueCopyvioUrls[] = $foundUrl;
-				}
-			}
+	public function getSources( $text ) {
+		// matches '[new line] ... (digits followed by %) (digits) ... (the URL)[word break]'
+		preg_match_all( '#\n\*.*?(\d+%)\s+(\d+).*?\b(https?://[^\s()<>]+)\b#', $text, $matches );
```
MusikAnimal: Here I simplified the old regex capturing the URL. It's unclear to what `[:punct:]` does or why we need it, but from my testing the above regex appears to work. Also I'm not sure what the comments `Determine if $value is a substring of an existing url, and if so, discard it` mean. I omitted this "hack" and all seems to work fine, but I'm probably wrong... are there any examples of where we need this?

PR #25, opened 2016-08-12
Filter out users in user whitelist, caching the whitelist in Redis

MusikAnimal:
Bug: T140204

### DIFF
--- a/src/Controllers/CopyPatrol.php
+++ b/src/Controllers/CopyPatrol.php
@@ -22,6 +22,7 @@
 namespace Plagiabot\Web\Controllers;
 
 use Wikimedia\Slimapp\Controller;
+use Wikimedia\Slimapp\Config;
 use GuzzleHttp;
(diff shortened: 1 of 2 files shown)


src/Controllers/CopyPatrol.php
```diff
@@ -290,6 +300,45 @@ protected function getFilterTypes() {
 	}
 
 	/**
+	 * Get cached user watchlist or re-fetch from wiki page and update redis
+	 *
+	 * @return array Usernames
+	 */
+	private function userWhitelist() {
```
kaldari: "userWhiteList" sounds like a member variable, not a function. Can we change this to getUserWhiteList() just to avoid any confusion? To quote https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP:
"Verb phrases are preferred: use getReturnText() instead of returnText()."

kaldari:
+2

PR #26, opened 2016-08-29
Add tests

Niharika29:
These tests don't work. Several issues:
* Loading the environment variables
* If I directly input the vars for testing, it gives me errors with Guzzle
* For the last test, the function doesn't return anything


Niharika29:
The Travis tests fail because the tests need access to environment variables to run which Travis doesn't know anything about. They work fine for me locally and should be good to run on staging/prod.

Niharika29:
Ouch. The environment variables work (via encryption) but Travis can't setup the SSH tunnels we need for getting the data. :(
I'm inclined to avoid running the tests during the build entirely and run them locally when needed.

MusikAnimal:
If we really wanted to, I _think_ we could do some magic in the `before_script` of the .travis.yml ([example](https://github.com/MusikAnimal/pageviews/blob/master/.travis.yml#L12-L15)) by piping in encrypted SSH keys, and then run the ssh command to set up the ports. Probably not worth the trouble, though :)

Niharika29:
After a lot of fighting with Travis, this finally seems good to merge. Tests will have to be run locally/on labs because Travis doesn't know how to setup ssh connections to tool labs dbs.

.travis.yml
```diff
@@ -1,12 +1,16 @@
...
+- '5.5'
+- '5.6'
+- '7.0'
+- hhvm
 sudo: false
 env:
   global:
-    - COMPOSER_DISABLE_XDEBUG_WARN=1
+  - COMPOSER_DISABLE_XDEBUG_WARN=1
+  - secure: YA404m0KGuqhJiq1fy6CtveOn37gC9yqqCVKUD+pdtshiUGq0CoEWqOfgpsktayCuRqsMsW/51SlKy7sOCHlpQZAz/pfkPKI/7x2Yb0oelDy/v7fPdOFdvZZkSdE/V5CLtKBgUWl5nuySANVg73uXj0MzRmhGAnSDWLZUD406TJUcTtypyVEUedUI5l8/92rjU/apNlrj+tgs3wu0KGNs7+JStBIS0jOkSWhYfgfwrNJTwlVFQ6xoX7a+NiCQx2U4NBwAF/I2+uyi7VvQg3txnf3lV2UluoL/5tq19qjIUSwjmlwhbOM6CAe26et5gJ6o9ACv/jFIw6pFzPuBAbwv0j5r/o2IpWo/p1XdlagMmBNGOdZXfFK9VuMXCXSj9hHiIx22kMNyWeI+F2s5gWtMbKVfC/K1svRj25ddEzdu+4Thn5Lr2DaIX5G1PAHn3Fl19d2Vn8ihE9rPY+gRpUJ0Ewt/YD32IiZPcgweVOuHIMTBq2+Yoc/30IKFM6tUhCPP7xldyYKgA1X1k4beyNyB/LBV060rn9R7I41Hht6sZlGZPMJm8zJTwCpzo7KX26FdIohBasSfDGCy4oecIkx6HimsB8meYkyJFElOWgSJs4BpiX0FL4kxxc8pqzki6wNpBPCfga5Inyw0HHKmuhprTV+5wR4uvPXKR78Y0uovxY=
```
kaldari: We don't need these now, right?

PR #27, opened 2016-09-15
Add i18n

Niharika29:
* Add SimpleI18n support to project
* Extract out all message keys to en.json
* Replace all keys in twig and js by message keys
* Update README


public_html/i18n/en.json
```diff
@@ -0,0 +1,70 @@
...
+  "imageattr-font": "<a target=\"_blank\" href=\"https://fonts.google.com/specimen/Bitter?query=bitter\">Bitter font</a> released under the <a target=\"_blank\" href=\"http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&amp;id=OFL_web\">OpenFont License </a>courtesy of <a target=\"_blank\" href=\"http://huertatipografica.com/\">Huerta Tipográfica</a>.",
+  "alert-testing": "This is a staging environment. Please do not use this tool to record your reviews. For the actual CopyPatrol tool, please use <a href=\"https://tools.wmflabs.org/copypatrol/\">tools.wmflabs.org/copypatrol</a>.",
+  "js-unauthorized": "You need to be logged in to be able to review.",
+  "js-dberror": "There was an error in connecting to database.",
+  "js-undo-own-only": "You can only undo your own reviews.",
+  "js-unknown-error": "An unknown error occurred when loading results. Please try again."
```
MusikAnimal: The only thing is I would avoid including links and markup in the messages. E.g. you get rid of `footer-source` and instead just have `"source": "Source"` and our HTML, `<a href=\"https://github.com/wikimedia/PlagiabotWeb\">{{ 'source'|message }}</a>`. The markup is not translatable and there's a chance someone might unintentionally break it.

And for `footer-poweredby` you could use: `Powered by $1 and $2` since EranBot and Turnitin are also not translatable. If they are, make a separate message for `turnitin` then in our HTML:
`{{ 'footer-poweredby'|message | ( <a href="https://en.wikipedia.org/wiki/Wikipedia:Turnitin">{{ 'turnitin'|message }}</a> | <a href="https://en.wikipedia.org/wiki/User:EranBot">{{ 'EranBot'|message }}</a> }}` (this is pseudocode, not sure how SimpleI18n works)

MusikAnimal:
Trying out these Gerrrit-wannabe features on GitHub, "Changes requested" doesn't mean I'm demanding any changes!

public_html/i18n/en.json
MusikAnimal: Also it should be `https://github.com/wikimedia/CopyPatrol`

Niharika29: Updated. I also fixed some links in Image attribution modal. I didn't break it up into 12 messages but 8 instead. 
The link in footer is <code>https://github.com/wikimedia/CopPatrol</code> only. Did you see otherwise anywhere?

Niharika29: Also, simpleI18n doesn't support the sort of markup you were suggesting above so I resorted to break out the text.

MusikAnimal:
I pushed directly to your branch, hope that's OK...

A few notes that I've learned doing i18n:

* For simple words or phrases, it's probably better to not make them specific to a particular part of the interface unless you are sure the message may differ from the identical translation based on where it's located. So if we used the word "WikiProjects" somewhere else, we wouldn't have to ask for a new translation but instead use the existing "wikiprojects" message (as opposed to "form-wikiprojects" and "col-wikiprojects").
* Avoid asking for translations of proper nouns. E.g. "EranBot" is a username and would be the same in every language
* Finally, we need to add qqq documentation, like [this](https://github.com/MusikAnimal/pageviews/blob/master/messages/qqq.json). For direct translations of words, such as "login", append "\n{{Identical}}" to the end of qqq message.

Sorry if I'm being anal! I just think we should try to make it easy as possible for the translatewiki folks, they are awesome.

Other not important things:
* I used spaces to separate the pipes in the twig templates. I found this easier to read, but purely a nitpick. Maybe I should find a Twig syntax highlighter.
* JavaScript variables are by convention camelCased
* For public-facing attribution I prefer to go by MusikAnimal :)

That is the extend of my stupid quibbles!

Niharika29:
> A few notes that I've learned doing i18n:
> 
> For simple words or phrases, it's probably better to not make them specific to a particular part of the interface unless you are sure the message may differ from the identical translation based on where it's located. So if we used the word "WikiProjects" somewhere else, we wouldn't have to ask for a new translation but instead use the existing "wikiprojects" message (as opposed to "form-wikiprojects" and "col-ewikiprojects")

For Copypatrol, I think I agree. When I did i18n for Wikimania scholarships app and Grants app, it was very prone to interface changes from time to time and hence separate keys were encouraged.
> Avoid asking for translations of proper nouns. E.g. "EranBot" is a username and would be the same in every language

Got it.
> Finally, we need to add qqq documentation, like this. For direct translations of words, such as "login", append "\n{{Identical}}" to the end of qqq message.

Thanks for reminding me to do this!
> Sorry if I'm being anal! I just think we should try to make it easy as possible for the translatewiki folks, because they are awesome.

Agreed. :)
> 
> Other not important things:
> 
> I used spaces to separate the pipes in the twig templates. I found this easier to read, but purely a nitpick. Maybe I should find a Twig syntax highlighter.

I've worked with twig before for quite a while ([here, for example](https://github.com/wikimedia/wikimedia-wikimania-scholarships/blob/master/data/templates/apply.html#L158)) and never used spaces with the pipe separator. If you don't mind, I'll like to continue to have it that way.
 
> JavaScript variables are by convention camelCased

Oh, right. 
> For public-facing attribution I prefer to go by MusikAnimal :).

Will remember in future.

public_html/i18n/en.json
```diff
@@ -0,0 +1,71 @@
...
+  "footer-loadmore": "Load More",
+  "footer-nomore": "No more records!",
+  "footer-broughtby": "Brought to you by $1",
+  "footer-commtech": "Community Tech",
+  "footer-poweredby": "Powered by $1 and $2",
+  "source": "Source",
+  "phabricator": "Phabricator",
+  "footer-imageattr": "Image attribution",
+  "lboard-title": "Leaderboard",
+  "lboard-week": "Cases closed in the<br/> last 7 days",
```
kaldari: Hardcoded linebreaks don't really work for translation, since the translators won't know where to put the linebreaks to make sure things wrap correctly. Danny originally wanted this done so that all the columns lined up right, but we can probably figure out a better way to do this in a future commit. For now, let's just remove the line breaks and let the lines wrap naturally (even if they don't all line up perfectly).

public_html/i18n/en.json
```diff
@@ -0,0 +1,71 @@
...
+  "login": "Login",
+  "logout": "Logout",
+  "header-loggedin": "You are logged in as: $1",
+  "form-all": "All cases",
+  "form-open": "Open cases",
+  "form-reviewed": "Reviewed cases",
+  "form-mine": "My reviews",
+  "form-drafts": "Drafts only",
+  "wikiprojects": "Wikiprojects",
+  "form-wikiprojects-placeholder": "Type Wikiproject names...",
```
kaldari: "WikiProject" is always camelcase.

Niharika29:
I've made the fixes and rebased the patch. I took this opportunity to split out the header, footer and modal templates from the base template. The code was nearly incomprehensible. 
I've let the twig pipe spaces be because I don't mind them if that's what MusikAnimal prefers. We're nearly done with this app anyway.

src/Controllers/CopyPatrol.php
```diff
@@ -150,31 +148,29 @@ protected function handleGet() {
...
-
 		// now all external requests and database queries (except
 		// WikiProjects) have been completed, let's loop through the records
 		// once more to build the complete dataset to be rendered into view
 		foreach ( $records as $key => $record ) {
-			$editor = $editors[$record['diff']];
+				$editor = null;
+				if ( isset( $record['diff'] ) && isset( $editors[$record['diff']] ) ) {
+					$editor = $editors[$record['diff']];
+				}
```
bd808: Block overindented?

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+      "Niharika Kohli",
+      "MusikAnimal",
+      "Ryan Kaldari"
+    ]
+  },
+  "name": "Title of the application/tool",
+  "documentation": "Link to documentation in header\n{{Identical|Documentation}}",
+  "feedback": "Link to feedback in header\n{{Identical|Feedback}}",
+  "leaderboard": "Link to leaderboard in header",
+  "login": "Butten label\n{{Identical|login}}",
```
bd808: "Button"

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+  "review-false": "Button label to mark a record as not fixed",
+  "review-false-tooltip": "Button tooltip to show help text",
+  "review-by": "Label to indicate which user reviewed a record",
+  "history": "Link to the history page for a given page title\n{{Identical|History}}",
+  "talk": "Link to the talk page for a given page title\n{{Identical|Talk}}",
+  "contribs": "Link to the contributions history for a given page title\n{{Identical|Contributions}}",
+  "record-editcount": "Label to indicate number of edits a user has made",
+  "compare": "Button label to open a panel for text comparison between two pages",
+  "record-ithenticate": "Button label for a button to go to the original plagiarism report generated by Turnitin",
+  "record-url-text": "Label text to show amount of plagiarism.\n Parameters: $1 - Percentage of edit that was plagirized.\n $2 - Number of words plagiarized.",
```
bd808: "plagiarized"

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+  "record-noeditor": "Column text shown to user when no editor was found in database for an edit",
+  "record-noeditor-tooltip": "Tooltip for a text label, shown to editor to explain why no editor could be found",
+  "record-ores": "Ores score for a record. \n Parameters: $1 - Decimal number indicating percentage score.",
+  "article": "Label shown for a text box, showing an article text. \n{{Identical|Article}}",
+  "compare-article-loading": "Temporary text shown to user while article text loads",
+  "compare-source-loading": "Temporary text shown to user while copyvio text source loads",
+  "footer-loadmore": "Label for a button to load more records.",
+  "footer-nomore": "Label to indicate no more records in database.",
+  "footer-broughtby": "Footer text to indicate authors. \n Parameters: $1 - String.",
```
bd808: $1 - List of names?

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+  "footer-commtech": "Footer text - name of team.",
+  "footer-poweredby": "Footer text to highlight sources used for development. \n Parameters: $1 - String. $2 - String.",
```
bd808: "String" is not descriptive at all for a translator.

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+  "lboard-month": "Column title for number of records closed last month by each person (top 10 only)",
+  "lboard-alltime": "Column title for the total number of records closed by each person (top 10 only)",
+  "lboard-back": "Button text to go back to CopyPatrol from leaderboard page",
+  "close": "Button text to close pop up box. \n{{Identical|Close}}",
+  "imageattr-title": "Title for popup box to show image attributions",
+  "imageattr-logo": "Title of logo, links to file.",
+  "public-domain": "Text as part of the attributions box",
+  "open-font-license": "Text as part of the attributions box",
+  "imageattr-paragraph1": "Text as part of the attributions box. \n Parameters: $1, $2, $3, $4 - Strings.",
+  "imageattr-paragraph2": "Text as part of the attributions box. \n Parameters: $1, $2, $3 - Strings.",
```
bd808: Again, the "Strings" description won't help a translator understand the resulting sentence/phrase.

MusikAnimal: We might want to make it clear the imageattr messages should be gender-neutral (User:Xander89 hasn't specified their gender, other proper nouns are not human) and that {{GENDER}} should not be used, since the parameters are links. In Intuition a PHP error is thrown if GENDER is used on markup, I think...

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
+{
+  "@metadata": {
+    "authors": [
+      "Niharika Kohli",
+      "MusikAnimal",
+      "Ryan Kaldari"
+    ]
+  },
+  "name": "Title of the application/tool",
```
kaldari: You may want to mark this one as optional to translate, since it's a proper noun and a translation isn't strictly needed. To mark something as optional to translate, add {{Optional}}\n to the beginning of it.

public_html/i18n/qqq.json
```diff
@@ -0,0 +1,71 @@
...
+  "documentation": "Link to documentation in header\n{{Identical|Documentation}}",
+  "feedback": "Link to feedback in header\n{{Identical|Feedback}}",
+  "leaderboard": "Link to leaderboard in header",
```
kaldari: "Leaderboard" may be tricky to understand. You might want to explain this better, like "Link to a list of the top reviewers in header".

PR #29, opened 2016-10-13
Generalised caching, more documentation, and other set-up things

samwilson:
This change adds a PSR-6 compatible caching system, instead of
only using Redis. This makes it easier to run locally (without
installing Redis).

Also extends the local-development setup instructions, making it
clearer what needs to be done to get things
up and running. Composer will now create a default .env file.

The Less cache is moved to the top-level 'cache' directory, because
this is where the cache is when Redis isn't used and because /src
seems like an odd place for volatile files. The new cache directory is added to the repo (empty) and a note about making it writable added to the readme.

The /login route is removed because it wasn't used anyway.

`composer.lock` is added, because then `composer install` will install the same thing for everyone, and the same in production (where `composer update` should never be run).

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -5,7 +5,7 @@ OAUTH_ENDPOINT       = 'https://meta.wikimedia.org/w/index.php?title=Special:OAu
 OAUTH_REDIR          = 'https://meta.org/wiki/Special:OAuth/authenticate?'
 OAUTH_CALLBACK       = 'https://tools.wmflabs.org/copypatrol/oauth/callback'
 
-; Database connection details. The same username and password is used for all three data sources.
+; Database connection details. The same username and password is used for both data sources.
 DB_DSN_ENWIKI      = "mysql:host=127.0.0.1;port=4711;dbname=enwiki_p"
 DB_DSN_PLAGIABOT   = "mysql:host=127.0.0.1;port=4711;dbname=s51306__copyright_p"
 DB_USER            =
--- a/.env_example
+++ b/.env_example
@@ -2,8 +2,8 @@
 OAUTH_CONSUMER_TOKEN =
 OAUTH_SECRET_TOKEN   =
 OAUTH_ENDPOINT       = 'https://meta.wikimedia.org/w/index.php?title=Special:OAuth'
-OAUTH_REDIR          = 'https://meta.org/wiki/Special:OAuth/authenticate?'
-OAUTH_CALLBACK       = 'https://tools.wmflabs.org/copypatrol/oauth/callback'
+OAUTH_REDIR          = 'https://meta.wikimedia.org/wiki/Special:OAuth/authenticate?'
+OAUTH_CALLBACK       = 'oob'
 
 ; Database connection details. The same username and password is used for both data sources.
 DB_DSN_ENWIKI      = "mysql:host=127.0.0.1;port=4711;dbname=enwiki_p"
--- a/.env_example
+++ b/.env_example
@@ -6,9 +6,8 @@ OAUTH_REDIR          = 'https://meta.org/wiki/Special:OAuth/authenticate?'
 OAUTH_CALLBACK       = 'https://tools.wmflabs.org/copypatrol/oauth/callback'
 
 ; Database connection details. The same username and password is used for all three data sources.
-DB_DSN_PLAGIABOT   = "mysql:host=localhost;dbname=copypatrol_dev"
-DB_DSN_WIKIPROJECT =
 DB_DSN_ENWIKI      = "mysql:host=127.0.0.1;port=4711;dbname=enwiki_p"
+DB_DSN_PLAGIABOT   = "mysql:host=127.0.0.1;port=4711;dbname=s51306__copyright_p"
 DB_USER            =
 DB_PASS            =
 
--- a/.env_example
+++ b/.env_example
@@ -1,0 +1,17 @@
+; Oauth parameters.
+OAUTH_CONSUMER_TOKEN =
+OAUTH_SECRET_TOKEN   =
+OAUTH_ENDPOINT       = 'https://meta.wikimedia.org/w/index.php?title=Special:OAuth'
+OAUTH_REDIR          = 'https://meta.org/wiki/Special:OAuth/authenticate?'
+OAUTH_CALLBACK       = 'https://tools.wmflabs.org/copypatrol/oauth/callback'
+
+; Database connection details. The same username and password is used for all three data sources.
+DB_DSN_PLAGIABOT   = "mysql:host=localhost;dbname=copypatrol_dev"
+DB_DSN_WIKIPROJECT =
+DB_DSN_ENWIKI      = "mysql:host=127.0.0.1;port=4711;dbname=enwiki_p"
+DB_USER            =
+DB_PASS            =
+
+; Caching system. If a Redis host is not defined, a local Filesystem cache will be used instead.
+REDIS_HOST = "tools-redis"
+REDIS_PORT = 6379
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -3,23 +3,41 @@ This is a web interface for [Plagiabot's Copyright RC feed](https://en.wikipedia
...
+   ```
+6. Open up an SSH tunnel to access the databases on Tool Labs (substitute your own username).
+   ```
+   $ ssh -L 4711:enwiki.labsdb:3306 YOU@tools-login.wmflabs.org -N 
+   ```
+7. Set up a local copy of the 'copyright' database (substitute your own usernames):
+   ```
+   $ mysqldump -u LABSYOU -p -P 4711 -h 127.0.0.1 --skip-lock-tables s51306__copyright_p > copyright.sql
+   $ mysql -u LOCALYOU -p -e "CREATE DATABASE copypatrol_dev;"
+   $ mysql -u LOCALYOU -p copypatrol_dev < copyright.sql
```
MusikAnimal: I would recommend connecting to the live copyright database through the same SSH tunnel, so in your `.env` you could use `DB_DSN_PLAGIABOT = 'mysql:host=127.0.0.1;dbname=s51306__copyright_p;port=4711'`

This way you don't need to recreate the data locally

composer.json
```diff
@@ -23,5 +25,13 @@
...
       "phpcs -p"
     ]
+  },
+  "script": {
+    "post-install-cmd": [
+      "php -r \"file_exists('.env') || copy('.env_example', '.env');\""
+    ],
+    "post-update-cmd": [
+      "php -r \"file_exists('.env') || copy('.env_example', '.env');\""
+    ]
```
MusikAnimal: super cool!

src/App.php
```diff
@@ -123,6 +125,24 @@ protected function configureIoc( \Slim\Helper\Set $container ) {
...
+							'port' => Config::getStr( 'REDIS_PORT' ),
+						],
+					]
+				] );
+			} else {
+				$driver = new FileSystem( [ 'path' => APP_ROOT . '/cache' ] );
+			}
+			$cache->setDriver( $driver );
+			return $cache;
+		} );
```
MusikAnimal: This is awesome! No disadvantage to having a local cache, but you could also make it work off the remote Redis via SSH tunnel `ssh -L 4711:enwiki.labsdb:3306 -L 4712:tools-redis:6379 YOU@tools-dev.wmflabs.org` then adjust `.env` accordingly. I should have added that to the README some time ago

README.md
samwilson: I thought about that, but doesn't it need to update it also? Perhaps I've read things wrongly... if so, then yeah definitely no need to recreate locally! :-) Much easier.

composer.json
samwilson: Yes, just a shame that there's no `post-update-or-install-cmd`! :-)

README.md
MusikAnimal: The `s51306__copyright_p` table is what's being continually updated by EranBot, so you'll have a continuous flow of new data, at least until people stop adding copyright violations :) It is true our users go through the backlog pretty quickly, so if you wanted to test something against a lot of unreviewed cases you probably won't be able to. For most dev work I don't think this will be a problem, though

samwilson: ha! no worries. :-)

MusikAnimal:
I'm not convinced the tests are working on Travis, unless it really takes 4+ hours... they pass locally so you've got my :+1:. Might need another set of eyes? Not sure

samwilson:
Yeah it's very strange. They took half an hour or so earlier, but reported themselves as 5 minutes or something.

I think the others should look at this first though. Just in case.

samwilson:
Travis must've just been busy doing other things; ey got it to eventually.

samwilson:
@Niharika29 would you mind casting your eye over this PR? Thanks!

.env_example
```diff
@@ -0,0 +1,16 @@
...
+
+; Database connection details. The same username and password is used for both data sources.
+DB_DSN_ENWIKI      = "mysql:host=127.0.0.1;port=4711;dbname=enwiki_p"
+DB_DSN_PLAGIABOT   = "mysql:host=127.0.0.1;port=4711;dbname=s51306__copyright_p"
+DB_USER            =
+DB_PASS            =
+
+; Caching system. If a Redis host is not defined, a local Filesystem cache will be used instead.
+REDIS_HOST = "tools-redis"
+REDIS_PORT = 6379
```
MusikAnimal: We might want to add `SLIM_MODE = 'development'` here, or at least I have on my local. Looking at the core Slim code I'm not convinced this doesn't do anything in particular by itself, but we might (or should) use it in other parts of the app, e.g. https://github.com/wikimedia/CopyPatrol/blob/master/src/Controllers/CopyPatrol.php#L364

.env_example
```diff
@@ -0,0 +1,16 @@
+; Oauth parameters.
+OAUTH_CONSUMER_TOKEN =
+OAUTH_SECRET_TOKEN   =
+OAUTH_ENDPOINT       = 'https://meta.wikimedia.org/w/index.php?title=Special:OAuth'
+OAUTH_REDIR          = 'https://meta.wikimedia.org/wiki/Special:OAuth/authenticate?'
+OAUTH_CALLBACK       = 'oob'
```
MusikAnimal: what is `oob`?

samwilson: 'Out Of Band', which isn't really correct because we do actually use a callback. (OOB usually means e.g. a desktop application is requesting the auth.) In our case, the callback is fixed by the consumer registration on Meta, so actually this will be ignored; it does have to be something though.

MusikAnimal: Interesting. Yeah for us it looks like we need a full URL to `/copypatrol/oauth/callback`. On my local I have the app running at `localhost/copypatrol` so I use `http://localhost/copypatrol/oauth/callback` and this works. We might suggest something similar here

MusikAnimal:
FYI I've got this branch up and running on our staging environment and all seems well :) https://tools.wmflabs.org/plagiabot

composer.json
```diff
@@ -1,11 +1,13 @@
 {
   "license": "GPL-3.0+",
   "require": {
+    "ext-intl": "*",
```
kaldari: What do we need the intl extension for?

MusikAnimal: That's for the number formatting I think. I did not know we could require PHP extensions through composer, so I ended up manually installing/enabling it my php.ini

samwilson: Yes, it's for NumberFormatter. Composer can't install extensions but it does alert people to the fact that they are required.

samwilson:
I closed and opened this earlier because the Github Android app is rubbish. :-)

composer.json
kaldari: Instead of requiring intl (which might require someone to spend half an hour upgrading their PHP), let's just check for the existence of NumberFormatter. After all, number formatting isn't critical to the tool's functionality:
```
if ( class_exists( 'NumberFormatter' ) ) {
	$formatter = new \NumberFormatter( $locale, \NumberFormatter::DECIMAL );
	// Get separator symbols (include decimal as it is required and we might use it at some point)
	$decimal = $formatter->getSymbol( \NumberFormatter::DECIMAL_SEPARATOR_SYMBOL );
	$thousands = $formatter->getSymbol( \NumberFormatter::GROUPING_SEPARATOR_SYMBOL );
} else {
	$decimal = '.';
	$thousands = ',';
}

Comment (samwilson) [inline composer.json]: Good idea! And we can make it simpler even, because the Twig defaults are the same as you've set there (dot and comma).

File: src/App.php [hunk 7]
```diff
@@ -179,6 +199,10 @@ protected function configureRoutes( \Slim\Slim $slim ) {
 				}
 			},
 			'twig-number-format' => function () use ( $slim ) {
+				if ( class_exists( \NumberFormatter::class ) ) {
```
kaldari: !class_exists

samwilson: Erk! Yes, well, hmm.... what can I say!?

:-)

(fixed now)

samwilson:
Yes, it's for NumberFormatter. Composer can't inatall extensions but itdoes alert prople to their requirement.


--

PR #32, opened 2016-10-21
Don't assume everything is English Wikipedia (T145436)

samwilson:
To resolve this task: https://phabricator.wikimedia.org/T145436


public_html/i18n/fr.json
```diff
@@ -0,0 +1,6 @@
+{
+	"name": "CopyPatrol",
+	"wikiprojects": "WikiProjets",
```
Niharika29: Wikiprojects? :)

samwilson: Yup! It's French: https://fr.wikipedia.org/wiki/Aide:Projet

But it's a good point: I didn't translate 'CopyPatrol'... is it "CopiePatrouille"? Or "patrouille du droit d'auteur" as google says?... :-) We need a French speaker here...

samwilson: Should we localise the logo? (The text bit of it, I mean.)

samwilson:
I've squashed all this into one commit, #34. Closing this one.

PR #33, opened 2016-10-21
Remove code assuming values in wikiprojects table are prefixed with WikiProject_

MusikAnimal:
Context: https://phabricator.wikimedia.org/T145436#2733046

I'm going to update MusikBot and the wikiprojects table to not include `WikiProject_` in the names, meaning we no longer need this bit of code.

Also, MusikBot ignores sub-WikiProjects (e.g. "Wikipedia:WikiProject Alternative music/The Smashing Pumpkins"), so that code here can also be removed.


MusikAnimal:
Going for it! Will migrate the wikiprojects data shortly

PR #34, opened 2016-11-03
Enable use on all Wikipedias

samwilson:
This change updates CopyPatrol to work on all language
Wikipedias that are supported by EranBot. Primarily it
focusses on getting things working with French (for
which translations are encluded here) but with the
exception of the WikiProject prefix cleanup in index.js
there's now nothing here specific to any one language.

Upgrading to this will require a couple of changes in
the .env file, and running `composer install`.

Bug: https://phabricator.wikimedia.org/T145436

This PR replaces #32 and squashes everything into one commit.


kaldari:
Hmm, the number formatting doesn't seem to be working in French. Currently, I see: "Compteur d'édition : 22,594", but I assume that should be "Compteur d'édition : 22 594".

samwilson:
Oops! It was because the Twig number formatting was being configured in route middleware that was never being loaded. I've moved it into `App::configureView()`.

samwilson:
Numbers display correctly now: http://tools.wmflabs.org/plagiabot/en/leaderboard?uselang=fr

Niharika29:
This looks great and works nicely. I'm going to merge this but let's not deploy it until we have Eranbot active on frwiki and musikbot configured all fine.

samwilson:
Sounds good. Thanks!

PR #38, opened 2016-11-17
Redirect to the root URL on logout

samwilson:
It was redirecting to the 'home' URL, which requires the language
parameter.

Bug: https://phabricator.wikimedia.org/T150919


samwilson:
This makes a double redirect, is that bad? It's simpler.

Niharika29:
Seems fine to me.

PR #40, opened 2016-11-21
Add a dropdown for switching between Wikipedias

samwilson:
This changes to using the Bootstrap navbar for the main menu, so
that it's easier to incorporate a dropdown with which to switch
from English to French or vice versa. This switch will to to the
corresponding other-language page (so if you're on the leaderboard
and switch you'll be on the other language's leaderboard).

Bug: https://phabricator.wikimedia.org/T150956

### DIFF
--- a/public_html/templates/header.html
+++ b/public_html/templates/header.html
@@ -1,16 +1,57 @@
 <header class="app-header">
-	<a href="{{ urlFor( 'home', {wikiLang:wikiLang} ) }}"><img src="{{ siteUrl( 'images/copypatrol.png' ) }}"/></a>
-	<a class="header-link" target="_blank" href="https://meta.wikimedia.org/wiki/CopyPatrol">{{ 'documentation' | message }}</a>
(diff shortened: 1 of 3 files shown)


public_html/templates/header.html
```diff
@@ -1,16 +1,60 @@
...
+								<span class="glyphicon glyphicon-log-out"></span> {{ 'logout' | message }}
+							</button>
+						</form>
+					</li>
+					{% else %}
+					<form action="{{ urlFor( 'oauth_init' ) }}" method="get">
+						<button class="btn btn-default navbar-btn">
+							<span class="glyphicon glyphicon-log-in"></span> {{ 'login' | message }}
+						</button>
+					</form>
```
MusikAnimal: Maybe make the `navbar-nav` a `<div>` or `<span>`, then under `{% if user %}` add the `<ul>` and `<li>`s, since `<form>` should not be a child of `<ul>`

samwilson: Good point! I've split it up, hopefully in a way that makes sense.

samwilson:
Changes made. https://tools.wmflabs.org/plagiabot/ updated to this branch. @MusikAnimal what do you think?

MusikAnimal:
:+1: Looks good to me!

PR #41, opened 2016-11-21
Fix issue with review buttons staying in loading state forever

Niharika29:


public_html/js/index.js
```diff
@@ -56,7 +56,7 @@
 			setReviewState( id, val );
 
 			$.ajax( {
-				url: baseUrl + 'review/add',
+				url: baseUrl + wikiLang + 'review/add',
```
Niharika29: Fixed.

PR #42, opened 2016-11-21
Get user whitelist to work again

Niharika29:
This commit:
* Removes Redis in favor of API call to get whitelist
* Handle error when whitelist page is empty

### DIFF
--- a/src/Controllers/CopyPatrol.php
+++ b/src/Controllers/CopyPatrol.php
@@ -107,7 +107,6 @@ class CopyPatrol extends Controller {
 		if ( empty( $records ) ) {
 			return $this->render( 'index.html' );
 		}
-		$userWhitelist = [];
 		$userWhitelist = $this->getUserWhitelist();
 		$diffIds = [];
 		$pageTitles = [];
--- a/src/Controllers/CopyPatrol.php
+++ b/src/Controllers/CopyPatrol.php
@@ -287,7 +287,7 @@ class CopyPatrol extends Controller {
 	}
 
 	/**
-	 * Get user watchlist from wiki page
+	 * Get cached user watchlist or re-fetch from wiki page and update redis
 	 *
 	 * @return array Usernames
 	 */
@@ -298,6 +298,7 @@ class CopyPatrol extends Controller {
 			return $whitelist;
 		}
 		$cacheKey = $this->wikiDao->getLang() . '_copypatrol_user_whitelist';
+		// Get whitelist from the cache if possible.
 		$cacheItem = $this->cache->getItem( $cacheKey );
 		if ( $cacheItem->isHit() ) {
 			$whitelist = $cacheItem->get( $cacheKey );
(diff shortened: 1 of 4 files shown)


src/Controllers/CopyPatrol.php
```diff
@@ -108,7 +108,7 @@ protected function handleGet() {
 			return $this->render( 'index.html' );
 		}
 		$userWhitelist = [];
```
MusikAnimal: don't need this anymore

src/Controllers/CopyPatrol.php
```diff
@@ -298,18 +298,8 @@ private function getUserWhitelist() {
...
-			$whitelist = $cacheItem->get( $cacheKey );
-		} else {
-			// It doesn't exist or it expired, so fetch from wiki page.
-			$whitelist = $this->wikiDao->getUserWhitelist();
-			// Store in the cache for 2 hours.
-			$cacheItem->set( $whitelist )->expiresAfter( 2 * 60 * 60 );
-			$this->cache->save( $cacheItem );
-		}
+		// It doesn't exist so fetch from wiki page.
+		$whitelist = $this->wikiDao->getUserWhitelist();
```
MusikAnimal: Let's try using the filesystem cache locally (remove redis vars from .env), and if that works for everyone maybe we could keep the caching functionality.

Niharika29: Yep, that worked fine.

src/Controllers/CopyPatrol.php
Niharika29: Removed.

MusikAnimal:
Looks good to me, but let's get #42 merged first since it's including in this. See my comment there

src/Controllers/CopyPatrol.php
```diff
@@ -288,7 +287,7 @@ protected function getFilterTypes() {
 	}
 
 	/**
-	 * Get cached user watchlist or re-fetch from wiki page and update redis
+	 * Get user watchlist from wiki page
```
MusikAnimal: restore original comment

src/Controllers/CopyPatrol.php
```diff
@@ -298,8 +297,7 @@ private function getUserWhitelist() {
 		if ( $whitelist !== null ) {
 			return $whitelist;
 		}
-		// Get whitelist from the cache if possible.
```
MusikAnimal: and maybe keep this comment, too

MusikAnimal:
Still confused why [this](https://github.com/wikimedia/CopyPatrol/pull/42/files#diff-d3bb9db39e409dbe4b7f9ce9de064a71R92) is showing up given it's already in master, but whatever, the code is there, so :+1: from me!

Niharika29:
Gonna squash commits. This is messy.

MusikAnimal:
One thing I didn't do was make sure this works when `Utilisateur:Example` is listed, as opposed to `User:Example`. I think it will

Niharika29:
I tested it. Works fine.

PR #43, opened 2016-11-23
Redirect to language version last used

Niharika29:


MusikAnimal:
:+1: Looks good to me! Tested it on my local

src/App.php
```diff
@@ -265,13 +265,21 @@ protected function configureRoutes( Slim $slim ) {
...
+					// See if we have a cookie indicating last version used
+					if ( isset( $_COOKIE['copypatrolLang'] ) ) {
+						$slim->redirectTo( 'home', [ 'wikiLang' => $_COOKIE['copypatrolLang'] ] );
+					} else {
+						// If no cookie, redirect to English version
+						$slim->redirectTo( 'home', [ 'wikiLang' => 'en' ] );
+					}
 					// Redirect root-URL requests to English.
 					$currentLang = $slim->i18nContext->getCurrentLanguage();
 					$slim->redirectTo( 'home', [ 'wikiLang' => $currentLang ] );
```
ricordisamoa: https://github.com/slimphp/Slim/blob/20a02782f76830b67ae56a5c08eb1f563c351a37/Slim/Slim.php#L1093
> This method immediately redirects to a new URL.

MusikAnimal: Is that not what we want?

ricordisamoa: I've not tested the code but it seems that the last `redirectTo` will never be called

MusikAnimal: Yep, I'm now realizing that as well. So I believe getting rid of the `else` block on lines 271-274 will do what we want. What do you think @Niharika29 ?

src/App.php
```diff
@@ -265,13 +265,21 @@ protected function configureRoutes( Slim $slim ) {
 				$middleware['inject-user'], $middleware['set-environment'],
 			function () use ( $slim, $middleware, $routeConditions ) {
 				$slim->get( '', function() use ( $slim ) {
+					// See if we have a cookie indicating last version used
+					if ( isset( $_COOKIE['copypatrolLang'] ) ) {
```
samwilson: It doesn't make much difference here, but there's `$slim->getCookie()` that will do this as well. It returns null if there's no cookie.

src/App.php
```diff
@@ -265,13 +265,21 @@ protected function configureRoutes( Slim $slim ) {
...
+						$slim->redirectTo( 'home', [ 'wikiLang' => 'en' ] );
+					}
 					// Redirect root-URL requests to English.
 					$currentLang = $slim->i18nContext->getCurrentLanguage();
 					$slim->redirectTo( 'home', [ 'wikiLang' => $currentLang ] );
 				} )->name( 'root' );
 				$slim->get( ':wikiLang',
 					function ( $wikiLang ) use ( $slim ) {
 						$page = new CopyPatrol( $slim );
+						setcookie( 'copypatrolLang', $wikiLang );
```
samwilson: Similarly to the above: `$slim->setCookie()` could also be used, I think.

src/App.php
Niharika29: Yeah, my mistake. Fixing that now.

src/App.php
Niharika29: I looked into that earlier. Slim only has session cookies whereas I thought we wanted ours to persist longer. I realize now I forgot to give it an expiry date. I'm thinking 30 days should be good.

src/App.php
```diff
@@ -265,13 +265,18 @@ protected function configureRoutes( Slim $slim ) {
 				$middleware['inject-user'], $middleware['set-environment'],
 			function () use ( $slim, $middleware, $routeConditions ) {
 				$slim->get( '', function() use ( $slim ) {
-					// Redirect root-URL requests to English.
-					$currentLang = $slim->i18nContext->getCurrentLanguage();
```
Niharika29: Makes sense.

src/App.php
```diff
@@ -265,13 +265,23 @@ protected function configureRoutes( Slim $slim ) {
...
+						$slim->redirectTo( 'home', [ 'wikiLang' => $_COOKIE['copypatrolLang'] ] );
+					} else {
+						// If no cookie, check if we support i18nContext's default language
+						$lang = $slim->i18nContext->getCurrentLanguage();
+						if ( in_array( $lang, [ 'en', 'fr' ] ) ) {
+							$slim->redirectTo( 'home', [ 'wikiLang' => $lang ] );
+						}
+						// We don't support i18nContext's current language, so redirect to en version
+						$slim->redirectTo( 'home', [ 'wikiLang' => 'en' ] );
+					}
```
Niharika29: It does look less understandable. I'd prefer sticking with the current version. :)

src/App.php
```diff
@@ -265,13 +265,23 @@ protected function configureRoutes( Slim $slim ) {
...
-					// Redirect root-URL requests to English.
-					$currentLang = $slim->i18nContext->getCurrentLanguage();
-					$slim->redirectTo( 'home', [ 'wikiLang' => $currentLang ] );
+					// See if we have a cookie indicating last version used
+					if ( isset( $_COOKIE['copypatrolLang'] ) ) {
+						$slim->redirectTo( 'home', [ 'wikiLang' => $_COOKIE['copypatrolLang'] ] );
+					} else {
+						// If no cookie, check if we support i18nContext's default language
+						$lang = $slim->i18nContext->getCurrentLanguage();
+						if ( in_array( $lang, [ 'en', 'fr' ] ) ) {
```
kaldari: Can we use a config var for this? i.e. $supportedLanguages? Or even just a member var at the top of the file would work. Something so it isn't as easy to miss when we expand the languages further.

src/App.php
```diff
@@ -49,6 +49,18 @@
 class App extends AbstractApp {
 
 	/**
+	 * @var array Languages supported by the app
+	 */
+	var $supportedLanguages = [ 'en', 'fr' ];
```
ricordisamoa:
```
 54 | ERROR | The var keyword must not be used to declare a property
 54 | ERROR | Visibility must be declared on property
    |       | "$supportedLanguages"
```

Niharika29: Thanks, I missed it. Fixed now.

src/App.php
```diff
@@ -49,6 +49,18 @@
 class App extends AbstractApp {
 
 	/**
+	 * @var array Languages supported by the app
+	 */
+	public $supportedLanguages = [ 'en', 'fr' ];
```
ricordisamoa: May I ask what is the purpose of having a public property and a getter at the same time. Thank you.

PR #44, opened 2016-12-02
Only show the WikiProject search field when required

samwilson:
This change hides the WikiProject search field if there
are no WikiProjects in the database for the Wikipedia language
in use.

Half of T152173 is fixed as well, because it's easier to test
this new functionality against a language (pt) that doesn't
have WikiProjects.

Composer is also updated, just for fun.

Bug: https://phabricator.wikimedia.org/T152134
Bug: https://phabricator.wikimedia.org/T152173

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -165,16 +165,16 @@
         },
         {
             "name": "guzzlehttp/promises",
-            "version": "1.2.0",
+            "version": "1.3.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/guzzle/promises.git",
-                "reference": "c10d860e2a9595f8883527fa0021c7da9e65f579"
+                "reference": "2693c101803ca78b27972d84081d027fca790a1e"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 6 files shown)


kaldari:
We should also hide the WikiProjects column in the results.

kaldari:
+1. This looks good to me. @MusikAnimal: Can you do a quick test of this version?

MusikAnimal:
It works but visually the column for WikiProjects is still there, it's just left blank. I know this is tricky to do but it'd be nice if we could hide it altogether so the content stretches across the full width of the screen, but what we have now is acceptable.

screen shot 2016-12-05 at 1 35 26 pm

kaldari:
Yeah, that's a bit awkward. Agree it would be better to remove it completely.

samwilson:
Good point. I was attempting to not make the code more complicated, but it does look odd. I've now stretched the 'Page', 'Editor', and 'Review' columns by one extra col each.

MusikAnimal:
Looks good! :+1: from me. That totals +2 and I think we're pretending GitHub is Gerrit so I have also merged

PR #45, opened 2016-12-02
Validate the wiki language in the route

samwilson:
Before, if an invalid language was requested (but one that still
passed the wikiLang regex, e.g. qq) the app would use it regardless.
This change adds a new middleware function that checks the incoming
language against the list of supported languages, and throws a 404
if it doesn't match.

Bug: https://phabricator.wikimedia.org/T152134

### DIFF
--- a/src/App.php
+++ b/src/App.php
@@ -272,79 +272,94 @@ class App extends AbstractApp {
 				}
 			},
 		];
-		$routeConditions = [
-			'wikiLang' => '([a-z]{1,3})'
-		];
-		$slim->group( '/', $middleware['trailing-slash'],
-				$middleware['inject-user'], $middleware['set-environment'],
-			function () use ( $slim, $middleware, $routeConditions ) {
-				$slim->get( '', function() use ( $slim ) {
-					// See if we have a cookie indicating last version used
-					if ( isset( $_COOKIE['copypatrolLang'] ) ) {
-						$slim->redirectTo( 'home', [ 'wikiLang' => $_COOKIE['copypatrolLang'] ] );
-					} else {
-						// If no cookie, check if we support i18nContext's default language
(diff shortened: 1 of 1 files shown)


src/App.php
```diff
@@ -311,12 +318,14 @@ function ( $wikiLang ) use ( $slim ) {
...
-					$page();
-				} )->name( 'loadmore' )->setConditions( $routeConditions );
+				$slim->get( ':wikiLang/loadmore', $middleware['valid-wikiLang'],
+					function ( $wikiLang ) use ( $slim ) {
+						$page = new Controllers\CopyPatrol( $slim );
+						$page->setDao( $this->getPlagiabotDao() );
+						$page->setWikiDao( $this->getWikiDao( $wikiLang ) );
+						$page();
+					}
+				)->name( 'loadmore' )->setConditions( $routeConditions );
```
MusikAnimal: Would it make sense to nest all of our `:wikiLang` routes under another group, and reference the middleware there? We have four such routes, so I'd argue the refactoring is worth it

MusikAnimal:
Just the question about adding a group for the wikiLang routes, otherwise looks good to me. I tested it on my local, works great

src/App.php
samwilson: Yeah, I actually refactored all the routes twice, before returning to the status quo; seemed simpler. But I'm absolutely more in favour of more-readable routes, so will have another crack at it if you think it's a good idea too. :-)

MusikAnimal: I trust your judgement! The extra group does mean another layer of nesting, which could be rough on the eyes. Up to you

samwilson:
I've rejigged the routes, and removed the middleware approach (because it was just doing a default 404 anyway) and instead have it constructing the wikiLang regex from the list of languages.

kaldari:
Looks OK to me. +2.

PR #46, opened 2016-12-05
Dynamically retrieve list of supported languages

samwilson:
The list of languages supported by CopyPatrol is retrieved from
the database, and cached for 7 days. The dropdown list in the
navbar now lists all these.

The phpcs change is just to ignore the cache directory that
exists for local development.

Bug: https://phabricator.wikimedia.org/T152362


src/App.php
```diff
@@ -49,15 +50,23 @@
...
+		$cacheItem = $this->slim->cache->getItem( 'supported-languages' );
+		if ( $cacheItem->isHit() ) {
+			return $cacheItem->get();
+		}
+
+		// Otherwise get the list from the database, and cache it for a week.
+		$langs = $this->getPlagiabotDao()->getLanguages();
+		$cacheItem->expiresAfter( new DateInterval( 'P7D' ) );
+		$cacheItem->set( $langs );
+		$this->slim->cache->save( $cacheItem );
```
MusikAnimal: Can't remember, where does the Slim cache get stored? In a cookie?

src/App.php
```diff
@@ -49,15 +50,23 @@
...
-		return $this->supportedLanguages;
+		// Use the cached list if possible.
+		$cacheItem = $this->slim->cache->getItem( 'supported-languages' );
+		if ( $cacheItem->isHit() ) {
+			return $cacheItem->get();
+		}
+
+		// Otherwise get the list from the database, and cache it for a week.
+		$langs = $this->getPlagiabotDao()->getLanguages();
+		$cacheItem->expiresAfter( new DateInterval( 'P7D' ) );
```
MusikAnimal: maybe also explain what `P7D` is. I'm safely guessing 7D = 7 days, but what does the P do?

src/App.php
```diff
@@ -203,7 +212,8 @@ protected function configureIoc( \Slim\Helper\Set $container ) {
 	protected function configureView( View $view ) {
 		$view->replace( [
 			'app' => $this->slim,
-			'i18nCtx' => $this->slim->i18nContext
+			'i18nCtx' => $this->slim->i18nContext,
+			'supportedLanguages' => $this->getSupportedLanguages(),
```
MusikAnimal: does the trailing comma violate any coding conventions (not sure)? I know some organizations prefer it this way

src/App.php
samwilson: No, Redis when in production, or a local file-based cache when in development (or any time we don't have `REDIS_HOST` defined).

src/App.php
samwilson: It's the [interval_spec](http://php.net/manual/en/dateinterval.construct.php), quite redundant really, I guess their aiming at future compatibility with things. I think it's also something to do with ISO8601 (with the 'T' being a delimiter for the time component)....

>  The format starts with the letter P, for "period." Each duration period is represented by an integer value followed by a period designator. If the duration contains time elements, that portion of the specification is preceded by the letter T. 

I'll make a comment about the format meaning.

src/App.php
samwilson: The mediawiki-codesniffer master doesn't complain about it, and I like doing trailing commas in multi-line arrays becuase it means the next time something gets added the diff then shows an actual meaningful change of 1 line.

MusikAnimal: Yeah I like it too :) As long as the codesniffer is happy, you don't want to upset it

samwilson: Talking of codesniffer, do you fancy reviewing this? https://gerrit.wikimedia.org/r/#/c/324376/ :-)

samwilson:
I've added messages for all languages supported by Turnitin (for the future).

src/App.php
kaldari: We don't have any coding conventions against trailing commas in PHP, in fact, this is probably the more common style in MediaWiki as it helps to prevent accidental syntax errors when adding new items to arrays. Of course, you have to be careful not to do this in JavaScript though.

PR #47, opened 2017-02-21
Add ability to link to a specific record via permalink

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T154811

This also fixes a bug where the WikiProject bubble links weren't functional on records shown with "Load more".

If you attempt to open a permalink to a record that is for a different project than what the cookie is set to, you are redirected and the cookie is changed.

### DIFF
--- a/public_html/templates/base.html
+++ b/public_html/templates/base.html
@@ -80,7 +80,7 @@
 
 	{% block main %}
 		{# Make the filters invisible but don't hide markup so that JS can react to clicking on WP bubbles #}
-		<section class="filters container-fluid {% if permalink %}none{% endif %}">
+		<section class="filters container-fluid{% if permalink %} none{% endif %}">
 			<form id="filters-form">
 				<fieldset class="form-group filters-fieldset text-center">
 					<div class="cases-filter col-sm-12">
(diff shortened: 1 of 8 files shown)


public_html/templates/base.html
```diff
@@ -79,7 +79,8 @@
 	{% endblock %}
 
 	{% block main %}
-		<section class="filters container-fluid">
+		{# Make the filters invisible but don't hide markup so that JS can react to clicking on WP bubbles #}
+		<section class="filters container-fluid {% if permalink %}none{% endif %}">
```
kaldari: Minor nitpick, but should probably be:
```
container-fluid{% if permalink %} none{% endif %}

File: public_html/templates/base.html [hunk 2]
```diff
@@ -143,10 +144,14 @@
 				{% endblock %}
 			</section>
 			<div class="load-more-div">
-				{% block loadmore %}
-					<button id="btn-load-more"
-					        class="btn btn-md btn-primary js-load-more">{{ 'footer-loadmore' | message }}</button>
-				{% endblock %}
+				{% if permalink %}
+					<a href="{{ baseUrl() }}" class="btn btn-md btn-primary btn-load-more">Show all records</a>
```
kaldari: "Show all records" hard-coded.

kaldari:
+2

PR #48, opened 2017-06-23
Degrade gracefully when ORES is down

eranroz:


eranroz:
PS: note that I already hotfixed the production with similar fix to make it work

MusikAnimal:
Thank you!

PR #49, opened 2017-10-12
Add search by page title

Niharika29:
* Keeping radio button for the possibility of expanding this to search by editor in future.
* Remove WikiProject search - not really used.
* Search by page title returns exact match + similar matches.


Niharika29:
I put this on our staging tool -
https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Tata+C&searchCriteria=page
https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Carl&searchCriteria=page
https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Snow&searchCriteria=page

src/Controllers/CopyPatrol.php
```diff
@@ -357,24 +357,22 @@ protected function getRecords() {
...
 		// make this easier when working locally
 		$numRecords = $_SERVER['HTTP_HOST'] === 'localhost' ? 10 : 50;
+
 		// compile all options in an array
 		$options = [
 			'filter' => $filter,
 			'last_id' => $lastId > 0 ? $lastId : null,
 			'drafts' => $drafts,
-			'wikiprojects' => $wikiprojects
+			'wikiprojects' => $wikiprojects,
```
MusikAnimal: Are we still supporting wikiprojects (outside showing them in the WikiProjects column)? I guess we could leave the code there just in case we want to revive it.

src/Dao/PlagiabotDao.php
```diff
@@ -68,6 +70,10 @@ public function getPlagiarismRecords( $n = 50, $options ) {
 					$filters[] = "status IS NULL";
 					break;
 			}
+			// search filters
+			if ( $searchCriteria == 'page' && $searchText ) {
+				$filters[] = "page_title LIKE '%$searchText%'";
```
MusikAnimal: I can't tell for sure but it looks like this should be a prepared statement, something like:
```php
$filters[] = "page_title LIKE '%:searchtext%'";
$preparedParams['searchtext'] = $searchText;
```

public_html/i18n/qqq.json
```diff
@@ -104,5 +104,7 @@
 	"js-undo-own-only": "Alert shown to user when they try to undo other people's reviews",
 	"js-unknown-error": "Alert shown to user when results could not be loaded, possibly due to connection issues, asking them to try again",
 	"permalink": "Text for a permanent link to a particular record.\n{{Identical|Permalink}}",
+	"search": "String label for a search box",
```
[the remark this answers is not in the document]
Niharika29: Good point.

public_html/templates/base.html
```diff
@@ -103,25 +103,25 @@
...
+						<div class="form-check">
+							<input type="radio" name="searchCriteria" value="page"
+									{% if searchCriteria == 'page' %} checked="checked" {% endif %}>
+							<label class="search-select-label" for="searchCriteria">
+								{{ 'search-page' | message }}
+							</label>
 
 						<button type="submit" class="btn btn-primary btn-sm btn-submit-filters">
 							{{ 'submit' | message }}
 						</button>
```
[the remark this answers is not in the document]
Niharika29: Whoops, yes!

public_html/templates/index.html
```diff
@@ -76,7 +76,7 @@
 					<div class="row-div col-sm-3 text-center wikiproject-div">
 						{% if row.wikiprojects is not empty %}
 							{% for wp in row.wikiprojects %}
-								<a class="row-div wproject" href="#">{{ wp }}</a>
+								<div class="row-div wproject">{{ wp }}</div>
```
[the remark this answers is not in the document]
Niharika29: You sure? I can't click on them locally or on the staging instance...

src/Dao/PlagiabotDao.php
Niharika29: Yeah, you're right. Been so long since I did this. 😞

public_html/templates/index.html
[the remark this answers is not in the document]
MusikAnimal: Yes, clicking on it just reloads the page. Looks like the JS is at: https://github.com/wikimedia/CopyPatrol/blob/master/public_html/js/index.js#L38-L42

public_html/i18n/qqq.json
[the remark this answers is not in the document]
Niharika29: Got rid of this string and replaced it with "Search for page title".

public_html/templates/index.html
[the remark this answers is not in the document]
Niharika29: Done.

Niharika29:
Plagiabot staging updated - https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Good&searchCriteria=page

src/Less/index.less
```diff
@@ -420,15 +425,14 @@ a[href=''] {
...
+	float: none;
+	margin-top: 2px;
+	padding: 2px;
 }
 
-.wikiproject-label {
-  display: block;
-  margin-top: 10px;
+.search-label {
+  padding-right: 1%;
```
kaldari: Labels are inline elements, so you should set it to "display: inline-block" to avoid any weirdness. (Technically, you're not supposed to apply padding to inline elements.) Also, using a percentage here is a bit awkward. We use pixel-based margins and paddings for most of the other elements in the header.

Niharika29: Done.

src/Less/index.less
```diff
@@ -431,8 +431,9 @@ a[href=''] {
 	padding: 2px;
 }
 
-.search-label {
-  padding-right: 1%;
+	.search-label {
```
kaldari: Looks great except for the extra tabs :)

src/Dao/PlagiabotDao.php
```diff
@@ -68,6 +67,11 @@ public function getPlagiarismRecords( $n = 50, $options ) {
 					$filters[] = "status IS NULL";
 					break;
 			}
+			// search filters
+			if ( $searchCriteria == 'page' && $searchText ) {
+				$searchText = filter_var( $searchText, FILTER_SANITIZE_STRING );
```
MusikAnimal: Going off of http://php.net/manual/en/filter.filters.sanitize.php it looks like `FILTER_SANITIZE_STRING` won't prevent SQL injection, instead just removing tags and encoding.

Did the bind keys thing not work? https://github.com/wikimedia/CopyPatrol/pull/49#discussion_r144409785

src/Dao/PlagiabotDao.php
```diff
@@ -38,20 +38,19 @@ class PlagiabotDao extends AbstractDao {
 	 *     'ithenticate_id'
 	 *   integer 'id' exact ithenticate_id of a record. This will override all
 	 *   	 other filter options
-	 *   string 'wikiprojects' pipe-separated list of wikiprojects
 	 *   string 'wikiLang' The language code of the Wikipedia to query for
```
MusikAnimal: `searchText` and `searchCriteria` should be added to this list. Also note the other options are underscored, e.g. `last_id` vs. `lastId`, `filter_user` vs. `filterUser`. Doesn't really matter, but might be good to be consistent.

src/Dao/PlagiabotDao.php
Niharika29: Yeah, bind doesn't work. According to docs, the placeholder cannot have quotes around it. It took me a while to figure out why everything was broken after I wrote that code and apparently this is somewhat obscure. But I cannot, for example, do "... LIKE '%:searchText%'"
I don't know what's a better way to do this. 🤔

Niharika29: I guess what we basically want is `mysql_real_escape_string()` but Slim abstracts it so we can't access the mysql connection object in order to be able to use that...

kaldari: Seems like you would want to use mysqli_real_escape_string() here instead of filter_var().

Niharika29: Can't - we don't have the connection object.
I think http://php.net/manual/en/function.mysql-escape-string.php will do the trick. It's deprecated (and later removed) but works for our case.

kaldari: @Niharika29: If you apply mysqli_real_escape_string() before it gets passed to Slim, I think it should work...
```
$searchText = mysqli_real_escape_string( $searchText );
$filters[] = "page_title LIKE '%" . $searchText . "%'";
```
Note mysqli rather than mysql, BTW.

MusikAnimal: I found this https://stackoverflow.com/a/12059603/604142

I would try something like:
```php
$filters[] = "page_title LIKE CONCAT('%', :searchtext, '%');
$preparedParams['searchtext'] = $searchText;
```

Niharika29: @kaldari - It won't. It needs a connection. See http://php.net/manual/en/mysqli.real-escape-string.php

@MusikAnimal - Tried that. Didn't work. I've killed quite some time around this bug now. :P

MusikAnimal: This might be a bug in Slim. `CONCAT` should work, e.g. try `CONCAT('%', 'Good', '%')`. I think it is inferring the wrong data type.

MusikAnimal: Nvm it is working for me... I tried this (also replacing spaces with underscores since that is how they are stored in the db):
```
$filters[] = "page_title LIKE CONCAT('%', :searchtext, '%')";
$preparedParams['searchtext'] = str_replace(' ', '_', $searchText);
```

src/Dao/PlagiabotDao.php
```diff
@@ -68,6 +69,11 @@ public function getPlagiarismRecords( $n = 50, $options ) {
 					$filters[] = "status IS NULL";
 					break;
 			}
+			// search filters
+			if ( $searchCriteria == 'page' && $searchText ) {
+				$searchText = mysql_escape_string( $searchText );
+				$filters[] = "page_title LIKE '%" . $searchText . "%'";
```
MusikAnimal: I would give this a try again, as prepared statements are much safer:
```php
$filters[] = "page_title LIKE CONCAT('%', :searchtext, '%')";
$preparedParams['searchtext'] = str_replace(' ', '_', $searchText);
```
This worked for me on my local.

The only other problem I noticed was after clicking on "Load more", the search text is no longer applied.

Niharika29:
Don't review yet.

MusikAnimal:
Test

MusikAnimal:
Test

Niharika29:
@kaldari Not yet. I'm yet to figure out why it's not taking the search text in consideration when I hit 'load more'. The code seems all fine. :/

Niharika29:
Never mind, I figured it out. Fix incoming.

Niharika29:
Live in action: https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=The&searchCriteria=page

Ready for review.

src/Less/index.less
```diff
@@ -420,15 +425,15 @@ a[href=''] {
...
+	margin-top: 2px;
+	padding: 2px;
 }
 
-.wikiproject-label {
-  display: block;
-  margin-top: 10px;
+	.search-label {
+		display: inline-block;
+		padding-right: 5px;
```
MusikAnimal: Indented one too many tabs?

Niharika29: Fixed!

public_html/templates/index.html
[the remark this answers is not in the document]
kaldari: If these aren't clickable, we need to change the text color from blue to black (or grey, or something besides blue).

src/Less/index.less
```diff
@@ -71,7 +71,7 @@ body {
 
 .wproject {
 	background: #f2f2f2;
-	color: #3f5fbf;
+	color: #2DABAD;
```
MusikAnimal: Maybe it's just me, but #2DABAD looks like a stark teal colour that doesn't fit in with the rest of interface. What if we removed the color property altogether and let it be the default #333 ?

kaldari: Sounds good to me.

PR #51, opened 2017-11-01
Set error output based on log level

samwilson:
Rather than have displayErrorDetails hard-coded to true.


MusikAnimal:
I've been wondering we were seeing full stack traces in production!

Now that we require `LOG_LEVEL` be set, I would uncomment that in .env_example, but this is fine for now.

PR #53, opened 2018-08-17
Allow searching for a page with the exact title

stephanebisson:
When searchCriteria == 'page_exact', only the records
for this exact page will be returned unlike 'page' that
does a SQL like '%text%' comparison.

This will be used by PageTriage to link to CopyPatrol
results for a specific page.

Bug: https://phabricator.wikimedia.org/T201075

### DIFF
--- a/src/Dao/PlagiabotDao.php
+++ b/src/Dao/PlagiabotDao.php
@@ -74,6 +74,10 @@ class PlagiabotDao extends AbstractDao {
 				$filters[] = "page_title LIKE CONCAT('%', :searchtext, '%')";
 				$preparedParams['searchtext'] = $searchText;
 			}
+			if ( $searchCriteria == 'page_exact' && $searchText ) {
+				$filters[] = "page_title = :searchtext";
+				$preparedParams['searchtext'] = $searchText;
+			}
 			// allow filtering by user and status
 			if ( $filterUser ) {
 				$filters[] = "status_user = '$filterUser'";


MusikAnimal:
Cool! :) Did we want to expose this option in the interface? I think it's actually OK to keep it as a hidden feature. Ping @Niharika29 

I actually can't test on my local because there are errors with PHP 7.2 compatibility (`Parameter must be an array or an object that implements Countable`). But I have this branch running on staging and it seems to work:
https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Choir+of+Somerville+College&searchCriteria=page
versus
https://tools.wmflabs.org/plagiabot/en?filter=all&searchText=Choir+of+Somerville+College&searchCriteria=page_exact (fuzzy search)

stephanebisson:
Thanks for the quick feedback on this. How would one go about deploying this?

MusikAnimal:
We can do that for you :) I wanted to get Niharika's feedback on whether we should expose this option in the UI, but we can decide that later. Merging and deploying now!

MusikAnimal:
Deployed https://tools.wmflabs.org/copypatrol/en?filter=all&searchText=Choir+of+Somerville+College&searchCriteria=page_exact

stephanebisson:
Great!

Thank you so much

Niharika29:
This would be a cool thing to expose in the UI but totally up to Growth team if they want to do this. :)
Thanks Leon and Stephane.

PR #54, opened 2018-09-30
Add noindex rule in page header

framawiki:
Bug: https://phabricator.wikimedia.org/T204369


MusikAnimal:
I don't think this will work for all web crawlers but it certainly helps, and is the more "proper" way to do it than blocking user agents. Thank you!

PR #55, opened 2018-11-01
SECURITY: Use prepared parameters consistently

catrope:
Don't rely on callers to escape/sanitize things.

Bug: T208280


MusikAnimal:
Merging given the urgency, but I personally am unable to test

PR #56, opened 2018-11-02
PlagiabotDao: Don't try to use typed prepared params

catrope:
Apparently they don't work and cause errors.


kostajh:
@catrope this line needs to be fixed as well https://github.com/wikimedia/CopyPatrol/blob/6d189480df7da55fe49a3560ad828ad9d1977e2e/src/Dao/PlagiabotDao.php#L62

catrope:
> @catrope this line needs to be fixed as well https://github.com/wikimedia/CopyPatrol/blob/6d189480df7da55fe49a3560ad828ad9d1977e2e/src/Dao/PlagiabotDao.php#L62

Well spotted! I've updated the PR to fix this.

kostajh:
LGTM

PR #57, opened 2018-11-02
Incorporate revision id search

kostajh:
This change allows one to pass in a `revision=:revId` parameter in the query
string. If present, we'll attempt to load results for that revision ID and will
merge them in with any other records found from the search.

Bug: T207345


kostajh:
Thanks @MusikAnimal. I staged the branch on plagiabot and left a comment here https://phabricator.wikimedia.org/T207345#4719044

MusikAnimal:
Seems to work! Merging

PR #58, opened 2019-02-17
Require PHP 7.2 and update dependencies

MusikAnimal:


MusikAnimal:
> Looks good to me. I couldn't actually test everything locally because (for some reason that I didn't look into too much) URL rewriting was not working.

Same! I ended up testing on https://tools.wmflabs.org/plagiabot. If I remember correctly this is where the Growth team was doing their development, too.

With your approval I'm going to merge

PR #90, opened 2020-10-08
Add endpoint to check if there's recent activity

MusikAnimal:
This is to monitor if EranBot is working properly. We'll use UptimeRobot
to continually ping the /activity_check endpoint, which will emaul us if
there is an unsuccessful response.

Bug: T262767


dayllanmaza (approved):
LGTM, seems to work fine against the staging environment.

samwilson (approved):
Looks good to me

PR #100, opened 2021-01-05
WikiDao: update to use new replicas

MusikAnimal:
Bug: T270407

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -13,6 +13,7 @@ DB_PASS            =
 DB_NAME_COPYPATROL = s51306__copyright_p
 
 ; Credentials for replicas.
+; On Toolforge, DB_REPLICA_HOST should be '*.web.db.svc.eqiad.wmflabs'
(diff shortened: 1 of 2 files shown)


MusikAnimal:
(copied from https://phabricator.wikimedia.org/T271186#6721729 ):

Here I did a cheap (but effective) solution of replacing a wildcard `*` with the correct database name when the connection is made. So, all we have to do is change the .env file on production to have `DB_REPLICA_HOST = *.web.db.svc.eqiad.wmflabs`. If the tool is being ran locally, you simply won't use a wildcard and instead hard-code your .env to connect to one of the databases.

This is live on staging: https://plagiabot.toolforge.org/ . The hostname is set to use the wildcard, so we know it's working because otherwise it'd error out given `*.web.db.svc.eqiad.wmflabs` is not valid.

samwilson (approved):
Works in my testing.

PR #132, opened 2022-06-03
Migrate earnbot's ithenticate script to CopyPatrol

MusikAnimal:
This means we no longer rely on an outside webserver to fetch document
IDs from Ithenticate.

Bug: T309816

### DIFF
--- a/src/Controllers/Ithenticate.php
+++ b/src/Controllers/Ithenticate.php
@@ -1,0 +1,75 @@
+<?php
+
+namespace Plagiabot\Web\Controllers;
+
+use Exception;
+use Slim\Slim;
+use Wikimedia\Slimapp\Controller;
+use PhpXmlRpc\Value;
+use PhpXmlRpc\Request;
+use PhpXmlRpc\Client;
+
+/**
+ * The Ithenticate controller receives the report ID from CopyPatrol,
+ * and redirects the user to the view-only URL of the Ithenticate report.
+ */
+class Ithenticate extends Controller {
+
+	/** @var Client */
+	protected $client;
+
+	/** @var string */
+	protected $rid;
+
+	/**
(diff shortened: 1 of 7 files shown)


src/Controllers/Ithenticate.php
```diff
@@ -0,0 +1,75 @@
...
+			'username' => $username,
+			'password' => $password,
+		] );
+		return $response->scalarval()['sid']->scalarval();
+	}
+
+	private function makeRequest( string $method, array $params ): Value {
+		$params = new Value( $params, 'struct' );
+		$request = new Request( $method, [ $params ] );
+		$response = $this->client->send( [ $request ] )[0];
```
samwilson, line 69: Will the result always be an array with an item? Might be worth guarding against it not being set.

samwilson (approved):
This all looks good to me! I haven't actually managed to run it all locally, and I'm not sure how I would given the requirement for an ithenticate account, but I'm sure it's fine.

My only thoughts are in handling responses from the API and not assuming their structure is always perfect, but I'm sure you've thought of that and it's all good.

src/Controllers/Ithenticate.php
MusikAnimal: It's an array if you give it an array of Requests, which begs the question of why we're sending an array of Request and not a single one since there's only ever one at a time. Fixed!

MusikAnimal:
Thanks for the review! I use the built-in PHP server locally, i.e. `cd public_html && php -S localhost:8000`. The JS/CSS assets aren't there, but the backend stuff works. One day we should make this easier to work on locally! iThenticate credentials will soon be in the .env file on Toolforge so you can find them there. With your OK I'm merging!

samwilson:
Great!

PR #144, opened 2022-11-11
i18n: Add simple english

theresnotime:
Add `simple-wikipedia` entries to i18n per step 3 from _README (Adding a new language)_, though I'm not entirely sure it's needed...

Bug: [T318032](https://phabricator.wikimedia.org/T318032)


MusikAnimal (approved):
Indeed it is needed to localize the label in the dropdown. My concern was the bot may not know how to handle the special lang code 'simple', but it most certainly did! https://copypatrol.toolforge.org/simple/?filter=all&searchText=&searchCriteria=page

The users have already cleaned out the backlog (there were only 10 reports).

Thank you for the PR! Merging

PR #158, opened 2023-03-14
Rename the CSS route 'styles' so that it loads natively on local envs

MusikAnimal:
Routes ending in .css or other common MIME types don't get loaded. There's probably a way to fix this for everyone, but simply not using .css as the extension will force it to work.

Other changes:
* Bump to PHP 7.4
* Use wikimedia/less.php, as oyejorge/less.php is abandoned
* Add ext-intl as a dependency in composer.json

Bug: [T332078](https://phabricator.wikimedia.org/T332078)
Bug: [T331220](https://phabricator.wikimedia.org/T331220)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,15 +1,16 @@
 {
   "license": "GPL-3.0-or-later",
   "require": {
-    "php": "^7.2",
+    "php": "^7.4",
     "ext-json": "*",
-    "oyejorge/less.php": "~1.5",
+    "ext-intl": "*",
     "mediawiki/oauthclient": "~1.0",
     "wikimedia/slimapp": "dev-master",
     "addwiki/mediawiki-api-base": "~2.1",
     "wikimedia/simplei18n": "~1.0",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -1,30 +1,29 @@
...
+    "phpxmlrpc/phpxmlrpc": "^4",
+    "wikimedia/less.php": "^4.0"
   },
   "autoload": {
     "psr-4": {
       "Plagiabot\\Web\\": "src/"
     }
   },
   "require-dev": {
-    "jakub-onderka/php-parallel-lint": "^1.0.0",
```
samwilson, line 20: The package is still supported and works. The old _name_ was abandoned, is all, and it's now called `php-parallel-lint/php-parallel-lint`.

samwilson (commented):
I've not actually tested, but this all looks good! (Shall test soon, just wanted to note this thing I saw.)

theresnotime (approved):
_hacky_ fix, but if it works.. 😄

PR #159, opened 2023-03-17
Add Dockerfile, development docker-compose

ChlodAlejandro:
This commit adds a Dockerfile for building a reproducible image of CopyPatrol based on Toolforge PHP 7.4 images. The Dockerfile mostly emulates the Toolforge environment, with slight modifications in development mode to make development easier (such as enabling XDebug in development mode).

The `docker-compose.dev.yml` file ensures that CopyPatrol is started with additional services, such as Redis, and mounts, such as the main program mount on the public user directory. It also ensures that port 9000/9003 are made accessible within the container to allow remote XDebug step debugging.

Note that the Docker Compose file binds to port 80 for the webserver. As the Twig templates are written now, ports are not properly handled when linking to server resources.

This adds the following checks:
* Build Docker image – performed on both `production` and `development` targets.
* Analyze Docker images – performs an efficiency check on the built Docker images using [Dive](https://github.com/wagoodman/dive).

### DIFF
--- a/.env_example
+++ b/.env_example
@@ -7,7 +7,7 @@ OAUTH_CALLBACK       = 'oob'
 
 # Credentials for private EranBot database.
 DB_HOST            = 127.0.0.1
-DB_PORT            = 4711
+DB_PORT            = 4712
 DB_USER            =
 DB_PASS            =
 DB_NAME_COPYPATROL = s51306__copyright_p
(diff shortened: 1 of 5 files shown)


theresnotime:
> Ah, right, I forgot about this. Will rebase when #160 gets merged. 😄

Done, go ahead :)

PR #157, opened 2023-03-05
Add option to search reviews by user

ChlodAlejandro:
**Note:** CI builds currently failing due to [T331220](https://phabricator.wikimedia.org/T331220).

This aims to fulfill a requested feature to search through reviews done by someone else.

* Searching by "reviewed cases" now shows a box for filtering by reviewer.
* The `filterUser` parameter can now be used to filter users.
* The leaderboard now includes a link to the user's reviews.
* Added an i18n message and its respective `qqq` entry.
* Converted the search area to CSS grid to make the new input look good.

Bug: [T331219](https://phabricator.wikimedia.org/T331219)

PR #163, opened 2023-03-20
Search interface improvements

ChlodAlejandro:
This makes a few interface improvements related to #157 and the search form.

* `reviewed` class converted to `data-filter` attribute. This means additional future search inputs can be controlled with similar CSS.
* The "Search reviews by user" field now appears as soon as "Reviewed cases" is selected as a filter.
* `search-form` class removed from the search form fieldset, the `filters-form` ID now handles that. Styles updated accordingly.
* The leaderboard reviews search link now has an appropriate `title` attribute.
* `search-page` i18n key renamed to `searchCriteria-page`. Since it looks like this key never really existed, this shouldn't be a problem. This resolves a missing key warning.
* `search` i18n key renamed to `filter-page` to remove ambiguity and match its variable name in code. This isn't mission-critical, and could be dropped should it be problematic for @translatewiki.


MusikAnimal (approved):
This looks and works great. Thank you kindly!

PR #15, opened 2016-06-23
Travis test fixup

bd808:
Make `composer test` run codesniffer tests and update Travis-CI config to run `composer test`. Branch includes a patch that fixes all current codeniffer warnings.

PR #168, opened 2023-07-13
Convert to Symfony and use ToolforgeBundle

MusikAnimal:
This is more or less a rewrite, but visually nothing should have changed. Moving to Symfony and using ToolforgeBundle makes the app more maintainable, especially for Community Tech (the maintaining team).

All controller actions are now in a single class, AppController. From here, we build Record objects that get pass to the views. Just as with before, there are separate classes to communicate with the db and APIs, now called CopyPatrolRepository and WikiRepository.

Summary of other changes:
* Add .editorconfig
* Use .env.local in local envs, with .env as the template
* Add .nvmrc
* Covert JS to ES6 and use Webpack Encore
* Remove old browser-prefixed rules from the LESS
* Rename status values to work with new bot/schema
* Update all PHP dependencies, remove those we don't need
* Remove code that now lives in ToolforgeBundle
* Remove unused messages from i18n
* Use NPM releases of bootstrap and jquery, and only include the parts that we need.
* Remove Select2 JS library as we're not using it anymore
* Do all validations server-side and not just in the JS
* Add caching to main DB queries and HTTP requests
* Convert the data structure for the view in to a Record model
* Fetch WikiProjects at the same time as redlink status of page titles
* Add stylelint and fix violations
* Add integration tests
* Other various code cleanup

This rewrite does NOT check against the user allowlist and auto-mark
records for users who are on it. That job is going to be moved to the
bot. See T341726.

Bug: T340600


MusikAnimal:
Depends on https://github.com/wikimedia/ToolforgeBundle/pull/59

I'm not sure what to do or how to fix the Docker issues.

samwilson:
It seems to be failing when running cache:clear in the composer script: `#15 5.254 !!  Could not open input file: ./bin/console`. Is it an issue with that relative path being called from somewhere other than the code directory?

ChlodAlejandro:
The Dockerfile (and other related Docker components) were added to ensure that the environment remains the same (since slight changes in environment made local development harder). With the Symfony migration, these don't really have a use anymore. They could be deleted and removed from the CI file.

The issue stems from how the Dockerfile is constructed, namely that packages are installed isolated from the rest of the repository using [multi-stage builds](https://docs.docker.com/build/building/multi-stage/). This helps decrease build times because of how Docker [caching](https://docs.docker.com/build/cache/) works. I know of a fix, but a decision on whether to keep the Docker things or not should probably go first. If that's a yes, a general makeover of the Dockerfile would also be needed to make it fit the new Symfony setup.

samwilson:
Removing the Docker setup I guess depends on who's working on CopyPatrol and how they want to work — it sounds like most people aren't using it? I must admit that I don't (but I haven't done anything with CopyPatrol for ages!).

If we want to look at updating it for Symfony, a somewhat similar Symfony app is WS Export, with this Docker config: https://github.com/wikimedia/ws-export/blob/main/docker/docker-compose.yml (not that I use that one eiter…).

MusikAnimal:
> Toolforge doesn't support custom containers

There's a new thing called [Build Service](https://wikitech.wikimedia.org/wiki/Help:Toolforge/Build_Service) that supposedly provides similar functionality, but I haven't tried it yet.

> With the Symfony migration, these don't really have a use anymore.

I don't understand what Symfony has to do with it. We're still going to be on Toolforge, if there was any confusion there. The Docker env doesn't have to use the Symfony server or CLI as the manual installation instructions imply.

I, too, don't personally use Docker for hardly anything other than MediaWiki, but I like having the option for folks who prefer it and gives them a lower barrier to contributing. That said, if no one wants to maintain it, we might as well remove it. I personally can't until I spend the time to master Docker... which I hope will happen some day... :)

Note there are Symfony recipes that create a Dockerfile for you ([docs](https://symfony.com/doc/current/setup/docker.html)). That actually happened when I first installed Symfony with this PR, but I reverted it back to the older Dockerfile so as not to lose that work, and I don't think the auto-generated Dockerfile worked out of the box anyway.

samwilson:
I'm happy to do the work to get Docker working with this PR, if that's helpful.

It sounds like it might be good to keep it, and if we stick with using it for CI (and use the toolforge images) then we have more assurance that the code will work on Toolforge (although, I don't think it's doing anything particularly out of the ordinary).

ChlodAlejandro:
After a few days of testing things, I think I've finally settled on a clean way to allow Docker-based development under the new setup. [Here](https://github.com/ChlodAlejandro/CopyPatrol/tree/new-backend-docker-3) are the changes.

The new setup requires that Symfony and PHP 7.4 be installed on the host machine to provide the ports to connect to Toolforge. Unfortunately, this doesn't fit well with the isolated environment that the containers are supposed to be in. So now, the SSH connection is instead made from within the development container. This does require that `$HOME/.ssh` be mounted on the container itself, but that shouldn't be any more of a security concern since it's no different to running everything on the host.

I've done my best to automatically configure the container using Docker Compose and a custom entrypoint script. This makes starting the whole thing just two commands (once the initial setup is done, of course). One of the advantages of this setup is that ports 4711 to 4720 are only accessible from within the container (but they could still be published using Docker). I've tested this so far on both Windows 11 and Ubuntu Focal (through GitHub Codespaces), and they both work great.

MusikAnimal:
I'm assuming @ChlodAlejandro can't push to this branch. Maybe if @samwilson's okay with everything else, we go ahead and merge then @ChlodAlejandro can make a new PR to fix the build? Or @ChlodAlejandro can make a new PR with my commit and their commit on top. I just want to make sure attribution is given where due :)

samwilson:
Yeah merging as-is sounds fine, and the fixing it after. @ChlodAlejandro are you happy with that?

ChlodAlejandro:
Sounds good! 😄 I'll file a PR for the Docker updates after this gets merged.

PR #169, opened 2023-07-25
Update Docker files (post-Symfony migration)

ChlodAlejandro:
This fixes the Docker configuration and Docker-related CI runs following the Symfony migration (#168).

* Moved Dockerfile to `docker/` folder
* Run post-install scripts later on in the build
* Move source file copy to end of target build
* Add OpenSSH, Symfony CLI, and Composer to development image
* Add entrypoint script (`/docker-entrypoint.sh`)
  * Add option to start Symfony local development server
  * Add option to start ToolforgeBundle SSH connection from image
* Optimize Dockerfile
* Add .dockerignore (copy of .gitignore)
* Fix docker-compose.dev.yml to conform to new project structure
  * Removed `copypatrol-redis` service
  * Automatically mount `$HOME/.ssh` to development container on startup
  * Change mounted folders to fit new structure
* Point ci.yml to new Dockerfile location
* Upgrade Docker-related CI steps


README.md
```diff
@@ -43,54 +43,76 @@ the [ToolforgeBundle](https://github.com/wikimedia/ToolforgeBundle).
...
+    5. `ITHENTICATE_USERNAME` and `ITHENTICATE_PASSWORD` are not necessary unless you need
        to test the "iThenticate report" functionality.
-4. Open up an SSH tunnel to access the databases on Toolforge. This assumes you have left
-   the `REPLICAS_HOST_*` and `REPLICAS_PORT_*` variables at their defaults.
+2. Build the development image once and install Composer dependencies with the following
    ```bash
-   symfony console toolforge:ssh --bind-address=0.0.0.0 --toolsdb
+   # (optional) Prevent double-downloading when the build occurs.
+   docker image pull docker-registry.tools.wmflabs.org/toolforge-php74-sssd-base:latest
+   docker-compose build
```
MusikAnimal, line 68: is the dash intentional? I've seen both, but on my machine I had to use `docker compose`

MusikAnimal (commented):
I'm assuming we need https://github.com/wikimedia/ToolforgeBundle/pull/61 before this will fully work?

Just in case it's unrelated: Following the directions in the README, everything worked great up until trying to open up the SSH tunnel. For me, it assumed `root@login.toolforge.org`. I'm not sure if that's expected. My local shell name is the same as on Toolforge, so as per the instructions I tried:
```
docker compose exec copypatrol start ssh musikanimal
```
but that gave me `Too many arguments to "toolforge:ssh" command, expected arguments "username".`

README.md
ChlodAlejandro: It seems `docker compose` is the new, suggested command. `docker-compose` refers to an old version of Docker Compose.<sup>[[1](https://stackoverflow.com/a/66516826/6011166)]</sup> This doesn't have much effect on Windows, but I guess that's not the case everywhere. I'll have this fixed in a later patch.

ChlodAlejandro:
> Following the directions in the README, everything worked great up until trying to open up the SSH tunnel.

wikimedia/ToolforgeBundle#61 *shouldn't be* required, but if I file a patch for the following, it then might be. The issue that [these lines](https://github.com/wikimedia/CopyPatrol/blob/5d70827c13192fea996803efe5bc136d276db27f/docker/docker-entrypoint.sh#L39-L42) attempt to solve is that SshCommand doesn't respect the `User` set in the user's `.ssh/config` file, and [applies the current username](https://github.com/wikimedia/ToolforgeBundle/blob/master/Command/SshCommand.php#L115C25-L115C25) regardless of that value (which would be `machine\user` on Windows systems). Submitting a username makes the command even longer than it currently is.

A user passing in their shell name *should* have been considered, but it seems that patch was lost, probably while I was (un)stashing around different solutions I had for the Docker fixes. The fix would be just avoiding the user of `$username` entirely when a variable without a dash is passed.

I could do the aforementioned, but I think it would be better to make ToolforgeBundle respect config-set `User`s instead, since this would have been expected behavior for someone with the value set in their SSH config. I didn't want to include this as part of wikimedia/ToolforgeBundle#61 though, since it would be a somewhat unexpected (if not breaking) change. Which solution would be better?

samwilson:
I've released a new version of ToolforgeBundle with those changes.

I like the idea of respecting the user's ssh config, if that's going to work. I guess there might be situations where someone hasn't configured ssh for Toolforge, but they'd likely be rare.

ChlodAlejandro:
Got it, I'll work on a patch for this in a bit. Thank you, @samwilson!

ChlodAlejandro:
Fixed the documentation issue and made `/app` read-write from within the container (so that Composer actions, etc. may be performed within the container environment). SSh issue is pending the result of wikimedia/ToolforgeBundle#63.

.dockerignore
```diff
@@ -0,0 +1,27 @@
...
+###> symfony/webpack-encore-bundle ###
+/node_modules/
+npm-debug.log
+yarn-error.log
+###< symfony/webpack-encore-bundle ###
+
+###> symfony/phpunit-bridge ###
+.phpunit.result.cache
+/phpunit.xml
+###< symfony/phpunit-bridge ###
```
MusikAnimal, line 27: Will this get updated by [Symfony Flex](https://symfony.com/components/Symfony%20Flex) like the normal .gitignore will? I'd just hate to have the two go out of sync. I assume this is to guard against committing things within the container. Would a symlink to the local .gitignore accomplish the same?

MusikAnimal (approved):
Works now in my testing! https://github.com/wikimedia/ToolforgeBundle/pull/63 was merged but a new release wasn't cut yet, so I tested with:
```
docker compose exec copypatrol symfony console toolforge:ssh --toolsdb -b 127.0.0.1 musikanimal
```
and it all worked flawlessly.

The possible issue with [.dockerignore](https://github.com/wikimedia/CopyPatrol/pull/169/files/5d192267326e047066306fb6fcc8f3b0274988ab#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5) becoming outdated can be addressed in a separate PR if desired.

Thanks so much for contributing!

.dockerignore
ChlodAlejandro: https://symfony.com/doc/current/setup/docker.html tells me that it *might* be, but I'm not so sure. Might be worth testing. But in any case, a symlink should work too.

ChlodAlejandro:
Looks like the change on ToolforgeBundle has been [tagged](https://github.com/wikimedia/ToolforgeBundle/releases/tag/1.5.2). Best to update the package soon so that the username issue won't occur down the line for other devs. Thank you both, @samwilson and @MusikAnimal!

PR #173, opened 2023-08-18
Always use submission IDs, and make other names more consistent

MusikAnimal:
This makes the frontend a tiny bit harder to understand, but stays consistent with the backend. Everything now uses submission IDs. The concept of 'state' is now the same as the report 'status'.

In turn, we now have working review buttons :)


MusikAnimal:
Somewhat urgent since we're asking folks to test out the new system, so self-merging. None of this is "production" per-se yet, anyway.

PR #174, opened 2023-08-29
Show edit size, summary, and tags for each edit

MusikAnimal:
Also cache the WikiRepository::draftsExist() query, which takes >100ms


MusikAnimal:
This can be tested on staging at https://plagiabot.toolforge.org

Apologies for the big PR. I'm mostly looking for a general go-ahead; no need to spend too much time reviewing unless you want to! This new version of CopyPatrol still isn't deployed, anyway, and won't be for probably another week or two.

src/Repository/WikiRepository.php
```diff
@@ -151,6 +151,124 @@ public function getLivePagesWithWikiProjects( array $titlesByNs ): array {
...
+			$cacheItem = $this->cache->getItem( $cacheKey );
+			foreach ( $response['query']['allmessages'] as $message ) {
+				$tagName = preg_replace( '/^Tag\-/', '', $message['name'] );
+				$messages[$tagName] = $message['content'] ?? $tagName;
+				// Fill in blank values for hidden tags.
+				$messages[$tagName] = $messages[$tagName] === '-' ? '' : $messages[$tagName];
+				// Cache for a week.
+				$cacheItem->set( $messages )
+					->expiresAfter( new DateInterval( 'P7D' ) );
+				$this->cache->save( $cacheItem );
```
samwilson, line 261: This overwrites the cached item on every iteration of the loop. Is that intentional?

MusikAnimal: No it is not! Thanks for catching this. Fixed :)

samwilson (approved):
I've not dug all that deeply, but I think this all looks good! :) Go for it, if you're happy.

PR #183, opened 2024-05-23
Set ACAO headers for public readonly API endpoints

sohomdatta1:
This will allow scripts, gadgets (and even potentially PageTriage) to surface richer information about the exact nature of the copyright violations to patrollers.


sohomdatta1:
Seems to already be active :( Not sure why it wasn't working.

PR #184, opened 2024-06-19
Update

Justman100:


Justman100:
@MusikAnimal

MusikAnimal:
Why are we removing the `revdel` message? I know it's not a real word, but it has a real meaning: https://www.mediawiki.org/wiki/Help:RevisionDelete

If anything, we could add `\n{{Optional}}` to the qqq documentation for the `revdel` message. I don't think there's a need to remove the ability to translate it entirely.

Justman100:
> Why are we removing the `revdel` message? I know it's not a real word, but it has a real meaning: https://www.mediawiki.org/wiki/Help:RevisionDelete
> 
> If anything, we could add `\n{{Optional}}` to the qqq documentation for the `revdel` message. I don't think there's a need to remove the ability to translate it entirely.

One cannot translate it! The way I see it, it's the abbreviation for RevisionDelete (RevDel)!

MusikAnimal:
> One cannot translate it! The way I see it, it's the abbreviation for RevisionDelete (RevDel)!

Correct. It's an intentionally short message, and it's up to the translator to come up with something that makes sense. If it nothing does, you can translate it literally, or just leave it as "revdel".

There are quite a few translations that exist already. For example, Russian uses "ревдел" which is also an abbreviation.

Justman100:
> > One cannot translate it! The way I see it, it's the abbreviation for RevisionDelete (RevDel)!
> 
> Correct. It's an intentionally short message, and it's up to the translator to come up with something that makes sense. If it nothing does, you can translate it literally, or just leave it as "revdel".
> 
> There are quite a few translations that exist already. For example, Russian uses "ревдел" which is also an abbreviation.

Then at least the sorting

Justman100:
@MusikAnimal

samwilson:
I don't think i18n files need to be sorted by key. They're usually ordered somewhat by software component, so that related messages are near each other regardless of their key's spelling.

PR #185, opened 2024-08-15
Process masked external links in tag wikitext

ChlodAlejandro:
Adds a `$includeExternalLinks` option to `parseWikitext` which enables parsing masked external links (`[https://w.wiki/M like this]`). This fixes tags showing up as raw wikitext (with a clickable link) instead of a properly-displaying tag. Edit summaries don't have this functionality, matching the behavior of MediaWiki.

This PR also adds additional tests and `rel="nofollow"` to external links.

n.b. This will make the function deviate from its [XTools implementation](https://github.com/x-tools/xtools/blob/4795fb88dd392bb0474219be3ef9a1fc019a228b/src/Model/Edit.php#L336). It might be good to change the code there as well.

Bug: [T372531](https://phabricator.wikimedia.org/T372531)

### DIFF
--- a/src/Model/Record.php
+++ b/src/Model/Record.php
@@ -192,7 +192,7 @@ class Record {
 	 */
 	public function getTagLabels(): array {
 		return array_map( function ( $tag ) {
-			return $this->parseWikitext( $tag );
+			return $this->parseWikitext( $tag, true );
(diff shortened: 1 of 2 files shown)


MusikAnimal (approved):
:100: 

This is fantastic. Thank you!!!

